PR #16, opened 2015-09-05
Remove pkg-/theme- from folder name

bpinto:
When installing packages (plugins) or themes using a URL, we should
strip the plugin-/pkg-/theme- from the repository name if it exists.

### DIFF
--- a/pkg/omf/cli/omf.install.fish
+++ b/pkg/omf/cli/omf.install.fish
@@ -36,7 +36,7 @@ function omf.install -a type_flag name_or_url
   if test -e $OMF_PATH/db/$parent_path/$name_or_url
     set target $parent_path/$name_or_url
   else
-    set -l local_name (basename $name_or_url)
+    set -l local_name (basename $name_or_url | sed "s/^pkg-//;s/^plugin-//;s/^theme-//")
     if test -e $OMF_PATH/$parent_path/$local_name
       echo (omf::err)"Error: $local_name $install_type already installed."(omf::off) 1^&2
     else


pkg/omf/cli/omf.install.fish
```diff
@@ -36,7 +36,7 @@ function omf.install -a type_flag name_or_url
   if test -e $OMF_PATH/db/$parent_path/$name_or_url
     set target $parent_path/$name_or_url
   else
-    set -l local_name (basename $name_or_url)
+    set -l local_name (basename $name_or_url | sed "s/pkg-//;s/plugin-//;s/theme-//")
```
bobthecow: 1. Anchor your regex, yo. 
2. This is a regex. You could totally do it with `(theme|pkg|plugin)` instead of three replacements.
3. Number 2 also keeps you from transforming my new `pkg-theme-plugin-maker` package folder into just `maker` :)

bpinto: I couldn't get `|` to work on a regex (unless using extended mode):

```
⋊> omf on correct-name-from-url ⨯ echo "theme-l" | sed "s/\(theme\|plugin\)-//"                                                                                           13:43:55
theme-l
⋊> omf on correct-name-from-url ⨯ echo "theme-l" | sed "s/\(theme|plugin\)-//"                                                                                            13:43:57
theme-l
⋊> omf on correct-name-from-url ⨯ echo "theme-l" | sed "s/(theme|plugin)-//"                                                                                           13:44:11
theme-l
```

bpinto: I have added the anchor element, but your example on number 3 is still not going to work. :+1: 

If you find a way to use `|` without using extended mode, let me know!

bobthecow:
any particular reason you don't want to use extended mode? Isn't it completely portable at this point?

bpinto:
Some plataforms use -r and some -E for the flag. Os I prefer to use basic sed to avoid platform specific bugs.

bobthecow:
basic regexen are _so bad_ though :(

on the two systems i tried (OS X and some recent Ubuntu) they both supported `-E` so as far as i'm concerned, we're good to go :)

bpinto:
> as far as i'm concerned, we're good to go :)

LMAO... if we were only concerned with those two OSs :P

bobthecow:
:+1:

scorphus:
Tested with OSX and Archlinux! Thanks :+1:

PR #19, opened 2015-09-06
Add grc plugin

tcoenraad:


bpinto:
Hey @tcoenraad, thanks for the PR. This is one of the steps that are necessary to add this plugin back to the repository.

But before we can merge, we need to do the proper changes to the `grc` plugin repository to adjust it to the new architecture. Similar to what was done here: https://github.com/oh-my-fish/plugin-sublime/pull/2/files

derekstavis:
The migration is done, awaiting reviews to merge.

PR #28, opened 2015-09-08
registry: add nvm plugin

derekstavis:


bpinto:
Unfortunately I don't think this is a good solution.

1. Plugin dependency does not handle order dependency. So a plugin might be installed but it's not yet as the plugin has not been loaded yet.

2. It tries to install nvm for you. We are not a package manager, Ta better to leave this to the user.

derekstavis:
I have pushed fixes for number 2. Number 1 is guaranteed as `available fenv` goes through the autoload paths and loads the given function. I also added cool completions now.

bpinto:
I don't understand the solution for number 1 being the available call. Could you test with a plugin that has a dependency from a plugin with a "higher" initial letter?

derekstavis:
I was wrong all the time. I've used `osx` plugin to test it and no way. It's a mere coincidence that plugins are loaded in a alphabetical order. But, well, the plugin works, and I think it's a great addition to omf, and we can continue this way while we dont tackle dependencies.

derekstavis:
@oh-my-fish/core any feedback on this?

bpinto:
:+1:

PR #37, opened 2015-09-10
Add more pkg to db

icyleaf:
Added two pkg:

1. git - mostly alias
2. xcode - functions
3. pod - competions

pls review, thanx

### DIFF
--- a/db/pkg/git
+++ b/db/pkg/git
@@ -1,0 +1,1 @@
+https://github.com/icyleaf/fish-pkg-git
--- a/db/pkg/pod
+++ b/db/pkg/pod
@@ -1,0 +1,1 @@
+https://github.com/icyleaf/fish-pkg-pod
--- a/db/pkg/xcode
+++ b/db/pkg/xcode
@@ -1,0 +1,1 @@
+https://github.com/icyleaf/fish-pkg-xcode


derekstavis:
Can you squash the "fix" commit and split this PR into two? Thanks!

icyleaf:
Done

derekstavis:
Great. Now, please, split the `Add pkg/xcode to db` commit into its own pull request :)

derekstavis:
We can accomodate a repository for you into the `oh-my-fish` organization. Would you like to?

bpinto:
@icyleaf the repository names need to become plugin-git plugin-xcode.

bpinto:
The git plugin only has aliases, I continue thinking we shouldn't add plugins with aliases only.

The Xcode plugin has every function defined on the same file. Could you split them into individual files?

icyleaf:
@bpinto  repository named must named plugin-git, plugin-xcode? in personal repository list, it will may other people lost it means. :(  i think @derekstavis way is great, thanks.

git plugin mostly are aliases and a little functions, the alias will help most develop guy in work life, are you agree?

i don't follow you means that "The Xcode plugin has every function defined on the same file. Could you split them into individual files", in omf repo, plugins and pkg is too few example.

BTW, what the different betwean pkg and plugins, i doubt ites.

icyleaf:
xcode plugin had been split each functions to it own files.

bpinto:
It's going to be easier to merge this if we have each plugin in a separate PR.

You might want to join our gitter (link on readme) as we have discussed having plugins with aliases and as everyone has a different alias, we think it's better not to have this sort of plugins.

icyleaf:
ok.

Issue #8, opened 2015-09-04
Plugin solution is not fishy


bpinto:
Instead of having every `function` define inside the `PACKAGE.fish` file, let's split the functions into individual files:

```
init.fish
uninstall.fish
PACKAGE.fish
```

It's not only a recommended way of doing it by [fish shell](http://fishshell.com/docs/current/tutorial.html#tut_autoload) but also more efficient because we won't load a function unnecessarily.

coderstephen:
That sounds cleaner to me. Reminds me of the old way of `*.load` files.

scorphus:
Absolutely, I was going to suggest this! It totally makes sense!

Gaelan:
:+1:

bobthecow:
if we're going this route, should't we move 'em to `/functions/*.fish` in the package directory?

bpinto:
@coderstephen 
Indeed it is, but changing the extension from `.load` to `.fish`. :+1: 

@bobthecow 
My initial thought was backwards compatibility, but it's doable.
Later I thought about most plugins that only have one function, and we would now have an extra directory for a single function.

Well, I think it's going to be well structured if we go this route. So, let's do it. The `init.fish` and `uninstall.fish` should not go into the `functions` dir, right?

derekstavis:
>Later I thought about most plugins that only have one function, and we would now have an extra directory for a single function.
I also think that we would like to implement a package description file (or function), where plugins can declare dependencies, so I think that this extra directory isn't a problem, it's a solution for the plugin infrastructure.

derekstavis:
Syntesizing what @bpinto is proposing + `package.fish` idea, a sample plugin directory structure would look like:

```
- plugin-foobar
  * init.fish
  * uninstall.fish
  * package.fish
  - functions
    *foo.fish
    *bar.fish
```

By only adding `functions` directory to autoloading paths we avoid loading unnecessary functions that are only used on load/init/uninstall.

bpinto:
That's it @derekstavis :+1:

derekstavis:
Hm, looks like it's identical, apart from just one file in `functions`

bpinto:
No `package.fish` file.

derekstavis:
Now that we revived this discussion, what about this:

- plugin-foobar
  * init.fish
  * uninstall.fish
  * package.fish
  - lib
    * foobar.fish
    * foo.fish
    * bar.fish

[...]

PR #43, opened 2015-09-11
bundle: make contents unique when adding and installing bundle

### LINKED ISSUE #42: Using omf install duplicates lines in bundle
Starting with no plugins installed and two things defined in the bundle file.

```
> cat .config/omf/bundle
theme bobthefish
package z
```

Now I'm running `omf install` to install my theme/packages.
```
> omf install
Installing bobthefish theme...
✔ theme bobthefish successfully installed.
Installing z package...
✔ package z successfully installed.
```

After that both things are getting added to the bundle (again)
```
> cat .config/omf/bundle
theme bobthefish
package z
theme bobthefish
package z
```

I'm using fish 2.2.0 and oh-my-fish 4b0dbc2

derekstavis:
This comes to fix #42 and also makes code more homogeneous by reusing code patterns.

### DIFF
--- a/pkg/omf/cli/omf.bundle.add.fish
+++ b/pkg/omf/cli/omf.bundle.add.fish
@@ -10,6 +10,8 @@ function omf.bundle.add -a type name_or_url
     else
       echo $record > $bundle
     end
+
+    sort -u $bundle -o $bundle
   end
 
   function __omf.write_theme
--- a/pkg/omf/cli/omf.bundle.install.fish
+++ b/pkg/omf/cli/omf.bundle.install.fish
@@ -23,6 +23,8 @@ function omf.bundle.install
         end
       end
     end
+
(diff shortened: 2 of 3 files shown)


bpinto:
What I don't like about sorting is that maybe a plugin requires another to be loaded before. I am not sure how to properly handle it yet. But it might be that the ordering should be irrelevant and we need to find a proper way to handle this scenario.

derekstavis:
Currently the bundle file isn't what is used to load packages. We need a dedicated solution for dependency tracking, and I think that it should be addressed together with the changes that #8 will implement, as this way we "break" the API in a row rather than sparse.

bobthecow:
:+1: looks good to me.

bpinto:
:+1:

derekstavis:
I've rebased the PR on master, merging it.

PR #49, opened 2015-09-13
Fix update when the pkg is not in database.

jeremiejig:
I change some behavior to avoid reloading when no installation has been done.

I also add a check so that pkg add by url (not present in `$OMF_PATH/db/pkg/`) will now update.

I previously add this behaviour when doing a `omf update`

```console
Updating Oh My Fish...
Oh My Fish is up to date.
Updating bobthefish theme...
✔ bobthefish theme up to date.
Updating apt package...
✔ apt package up to date.
Updating bundler package...
✔ bundler package up to date.
Updating extract package...
✔ extract package up to date.
Error: mc package already installed.
```

### DIFF
--- a/pkg/omf/cli/omf.install_package.fish
+++ b/pkg/omf/cli/omf.install_package.fish
@@ -1,10 +1,10 @@
 function omf.install_package -d "Return success for at least one successful install."
-  set -l return_code 1
+  set -l return_code $OMF_UNKNOWN_ERR
   for search in $argv
     omf.install --pkg $search
-    set -l code $status
-    if test $code -eq 0
-      set return_code $code
+    set -l install_code $status
+    if test $install_code -eq 0
+      set return_code $install_code
     end
   end
   return $return_code
--- a/pkg/omf/cli/omf.install_package.fish
+++ b/pkg/omf/cli/omf.install_package.fish
@@ -1,4 +1,5 @@
-function omf.install_package -d "Return success for at least one successful install."
+# Return success for at least one successful install.
+function omf.install_package 
   set -l return_code $OMF_UNKNOWN_ERR
   for search in $argv
     omf.install --pkg $search
--- a/pkg/omf/cli/omf.install_package.fish
+++ b/pkg/omf/cli/omf.install_package.fish
@@ -1,5 +1,11 @@
-function omf.install_package
+function omf.install_package -d "Return success for at least one successful install."
+  set -l return_code 1
   for search in $argv
     omf.install --pkg $search
+    set -l code $status
+    if test $code -eq 0
+      set return_code $code
+    end
   end
+  return $return_code
 end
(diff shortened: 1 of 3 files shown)


pkg/omf/cli/omf.install_package.fish
```diff
@@ -1,5 +1,11 @@
-function omf.install_package
+function omf.install_package -d "Return success for at least one successful install."
+  set -l return_code 1
   for search in $argv
     omf.install --pkg $search
+    set -l code $status
+    if test $code -eq 0
```
bpinto: Shouldn't this check return_code instead?

pkg/omf/cli/omf.install_package.fish
```diff
@@ -1,5 +1,11 @@
-function omf.install_package
+function omf.install_package -d "Return success for at least one successful install."
+  set -l return_code 1
```
bpinto: We define some return code on omf.fish, better to use one of those.

bpinto: As you did above.

pkg/omf/cli/omf.install_package.fish
jeremiejig: No because as I had in the description I want to return 0 when at least one omf.install succeed.
I'll push another commit to make it clearer.

pkg/omf/cli/omf.install_package.fish
```diff
@@ -1,5 +1,11 @@
-function omf.install_package
+function omf.install_package -d "Return success for at least one successful install."
```
bpinto: This is not a package description.

bpinto: Function description

pkg/omf/cli/omf.install_package.fish
bpinto: Maybe do the same you did on remove and only install one package at a time by adding the loop to the function caller?

pkg/omf/cli/omf.install_package.fish
```diff
@@ -1,5 +1,12 @@
-function omf.install_package
+# Return success for at least one successful install.
+function omf.install_package 
+  set -l return_code $OMF_UNKNOWN_ERR
   for search in $argv
     omf.install --pkg $search
+    set -l install_code $status
+    if test $install_code -eq 0
```
bpinto: I still think this is unexpected.

I have worked on a bigger change for the install command but didn't have the chance to complete it. You might want to use it as a starting point?

https://github.com/oh-my-fish/oh-my-fish/commit/26c3b722711cc98688e63c6b39b267f5f676a235

jeremiejig: At first read, this make theme pkg mandatory in case of name collision.
I don't know if it is a expected behaviour.

It does give me some idea for the organization of my code.

bpinto: Yep, we are not really worried about naming collision as the prefix are recommended/mandatory.

jeremiejig: So, I'm restarting from your commit. I saw that you made the install command to do only an installation (quite logical).

So I have to create an omf.update similar to the install ?

For now I will continue with two command :omf.install and omf.update.

jeremiejig:
I'm closing this PR, will be fixed by #49.

bpinto:
Thanks!

PR #50, opened 2015-09-13
Migrate old config format to bundle

derekstavis:
This PR makes `omf install` without parameters always check old config `~/.config/fish/config.fish` and migrate `Plugin` and `Theme` entries to new bundle format. It's not the perfect approach to the matter, but it works as a palliative for the time being.

### DIFF
--- a/pkg/omf/util/omf.util.migrate_old_bundle.fish
+++ b/pkg/omf/util/omf.util.migrate_old_bundle.fish
@@ -1,0 +1,22 @@
+function omf.util.migrate_old_bundle
+  set OMF_OLD_CONFIG ~/.config/fish/config.fish
+  set OLD_PLUGIN (grep "Plugin" $OMF_OLD_CONFIG 2> /dev/null)
+  set OLD_THEMES (grep "Theme" $OMF_OLD_CONFIG 2> /dev/null)
+  set BUNDLE_PATH $OMF_CONFIG/bundle
+
+  if test -z "$OLD_PLUGIN" -a -z "$OLD_THEMES"
+    return 0
+  end
+
+  for theme in $OLD_THEMES
+    omf.bundle.add theme (echo $theme | sed -n 's/Theme "\(.*\)"/\1/p')
+  end
+
+  for plugin in $OLD_PLUGIN
+    omf.bundle.add package (echo $plugin | sed -n 's/Plugin "\(.*\)"/\1/p')
+  end
+
+  sed -i '/Theme.*/d;/Plugin.*/d' $OMF_OLD_CONFIG
+
+  return 0
+end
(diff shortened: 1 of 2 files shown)


bpinto:
I'm not sold yet that we should remove the old syntax. So I'd prefer a simpler solution that would alias both Plugin and Theme to omf.bundler.add

pkg/omf/util/omf.util.migrate_old_bundle.fish
```diff
@@ -0,0 +1,22 @@
+function omf.util.migrate_old_bundle
+  set OMF_OLD_CONFIG ~/.config/fish/config.fish
+  set OLD_PLUGIN (grep "Plugin" $OMF_OLD_CONFIG 2> /dev/null)
+  set OLD_THEMES (grep "Theme" $OMF_OLD_CONFIG 2> /dev/null)
+  set BUNDLE_PATH $OMF_CONFIG/bundle
+
+  if test -z "$OLD_PLUGIN" -o -z "$OLD_THEMES"
```
bpinto: Shouldn't this be a -a?

pkg/omf/util/omf.util.migrate_old_bundle.fish
```diff
@@ -0,0 +1,22 @@
...
+    return 0
+  end
+
+  for theme in $OLD_THEMES
+    echo $theme | sed 's/Theme/theme/g;s/["|\']//g' >> $BUNDLE_PATH
```
bpinto: Use omf.bundle.add to avoid duplicating items?

derekstavis: Good idea!!!

pkg/omf/util/omf.util.migrate_old_bundle.fish
derekstavis: Sure. Thanks!

derekstavis:
Aliasing functions is also a good idea, I've tried this approach, as we avoid errors on the first run, but remember that `config.fish` sources `init.fish` before the `Plugin` and `Theme` directives, so `omf.bundle.add` isn't in autoload paths. We would need to change the startup process.

bpinto:
True. If only we had a way to handle dependencies between plugins, this would work.

derekstavis:
@oh-my-fish/core reviews? ideas? or we forget this and close it?

bpinto:
I think we can make this easier. I'd like to try cloning the functions for now.

bpinto:
@derekstavis What do you think?

derekstavis:
I think it's not a complete solution, if I understand what you mean. It tackles backward compatibility, but doesn't migrate old configuration scheme to new configuration scheme. Perhaps, could you explain what you have in mind?

bpinto:
I'll try coding something soon.

PR #51, opened 2015-09-15
Merge theme and install, create update function. (WIP: remove all `cd`)

jeremiejig:
Hi,

* So starting from the work (26c3b72) of @bpinto with a rebase, I'd like to propose this PR which will also fix #49.
This is still a WIP, as I haven't done any testing. I would like some code review.

* I will remove all use of `pushd`, `popd` and `cd` when doing some `git` command. see [`git -C`](https://github.com/git/git/blob/v2.5.2/Documentation/git.txt#L486)

### DIFF
--- a/pkg/omf/cli/omf.update.fish
+++ b/pkg/omf/cli/omf.update.fish
@@ -1,0 +1,42 @@
+function omf.update -a type_flag name
+  function __omf.update.process
+    echo (omf::em)"Updating $argv..."(omf::off)
+  end
+
+  function __omf.update.success
+    echo (omf::dim)"✔ $argv up to date."(omf::off)
+  end
+
+  function __omf.update.error
+    echo (omf::err)"Could not update $argv."(omf::off) 1^&2
+    return $OMF_UNKNOWN_ERR
+  end
+
+  function __omf.update.not_installed
+    echo (omf::err)"Could not find $argv."(omf::off) 1^&2
+    return $OMF_UNKNOWN_ERR
+  end
+
+  switch $type_flag
+    case --theme
+      set update_type "theme"
+      set parent_path "themes"
+    case --pkg
+      set update_type "package"
+      set parent_path "pkg"
+    case *
+      echo (omf::err)"Argument to omf.update must be --theme name or --pkg name"(omf::off)
+      return $OMF_INVALID_ARG
+  end
+
+  if test -d $OMF_PATH/$parent_path/$name
+    __omf.update.process $update_type $name
+    pushd $OMF_PATH/$parent_path/$name
+    omf.repo.pull
+    popd
+    __omf.update.success $update_type $name
+  else
+    __omf.update.not_installed $update_type $name
+  end
+
+end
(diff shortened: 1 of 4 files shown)


derekstavis:
I don't like having `--type`. What we need is to guarantee that no package name clashes occur.

jeremiejig:
I can remove the --type and add a some check that there isn't a theme or plugin with the same name already installed.

derekstavis:
I think we don't need such checks. The gist is to merge `db/pkg` and `db/theme` directories, this way we avoid duplication in the filesystem level. Is this the initial idea @bpinto?

jeremiejig:
So is there a policy that forbid name clash between `plugin-` and `theme-` ?

If we merge `db/pkg` and `db/theme`, we change the name of the files (and folder) from `myplugin` to `plugin-myplugin` and `theme-awesome`

If this is the goal (avoid duplication of folder), I can continue to work on it.
I just need a confirmation on the guideline.

bpinto:
We could merge db/pkg and db/theme. I'm open to this although it was not going in this direction. There is no rule that forbids. It just never happened and although rare, if it happens we can ask for a rename.

derekstavis:
If we merge packages and plugins `--type` isn't needed anymore.

bpinto:
My initial commit was checking the database to see if you wanted to install a plugin or theme.

When you mean merge, what exactly do you mean with that?

derekstavis:
"Everything is a package". Examples:

    omf install agnoster
    omf install title
    omf install bobthefish
    omf install rbenv

bpinto:
Right, but would we keep the prefixes? They are important.

derekstavis:
I think yes, because we need a way to distinguish what is a theme and what is a plugin.

bpinto:
We are in sync then :)

jeremiejig:
Mmmh I think I understand the architecture.

So I'll move everything into one folder.
```
db/
pkg/
```

bpinto:
@jeremiejig we are back in game! Let me know if you need help here!

pkg/omf/cli/omf.update.fish
```diff
@@ -0,0 +1,42 @@
...
+    echo (omf::err)"Could not update $argv."(omf::off) 1^&2
+    return $OMF_UNKNOWN_ERR
+  end
+
+  function __omf.update.not_installed
+    echo (omf::err)"Could not find $argv."(omf::off) 1^&2
+    return $OMF_UNKNOWN_ERR
+  end
+
+  switch $type_flag
```
bpinto: You could use the same logic applied on `omf.install` to remove this type_flag.

pkg/omf/cli/omf.update.fish
```diff
@@ -0,0 +1,42 @@
...
+      set update_type "package"
+      set parent_path "pkg"
+    case *
+      echo (omf::err)"Argument to omf.update must be --theme name or --pkg name"(omf::off)
+      return $OMF_INVALID_ARG
+  end
+
+  if test -d $OMF_PATH/$parent_path/$name
+    __omf.update.process $update_type $name
+    pushd $OMF_PATH/$parent_path/$name
```
bpinto: `pushd` and `popd` are not necessary. :D
`omf.repo.pull $OMF_PATH/$parent_path/$name` instead.

bpinto:
Completed on #121

PR #52, opened 2015-09-17
Travis: Base of travis test

jeremiejig:
Hi,

In this PR, I've set up the environment of .travis.yml here are the main point:

* Setting up `OMF_REPO_URI` and `OMF_REPO_BRANCH` with Travis variable
* Building and caching fish-2.2.0 to be able to run some more test on omf :sparkles: 
* Adding an `tests/travis_install` to catch git clone to build the correct commit on PR (never tested before this PR will trigger the first test !).
* adding a folder `tests/fish_user_function` to override some function like refresh (if not travis build will hang).
* `omf.repo.pull [repo_dir]` to remove pushd and popd, because I think than apart from omf cd, we do not want to mess up the dir stack and dir history of the user.

Happy reviewing :smile:


derekstavis:
You can squash  e3bb044 and e3f08ca, as one edits the description and the other remote it.

derekstavis:
Thanks for the changes!! I haven't tested them yet, but I think that there's some stuff out of this PR scope. The commits 8a97d4b, e3bb044 and e3f08ca should be moved to another PR, which me can merge faster and before this one to make the process smoother.

jeremiejig:
Well I just squashed the two last commit into 8a97d4b see 1f6f0f8.

* about the omf.repo.pull, I kind of push it with this PR because it does use the same logic has `bin/install`

* `bin/install` was broken because `pushd` `popd` are not implemented in `sh` (at least on Debian).

.travis.yml
```diff
@@ -8,15 +8,36 @@ sudo: false
...
+
+env:
+  global:
+    - PATH="$HOME/fish/bin:$PATH"
+    - OMF_REPO_URI="https://github.com/$TRAVIS_REPO_SLUG"
+    - OMF_REPO_BRANCH="$TRAVIS_BRANCH"
+
+before_install:
+  - tests/install_fish.sh
+  - fish -c 'ln -sf $TRAVIS_BUILD_DIR/tests/fish_user_function $fish_function_path[1]'
```
bpinto: What is this?

.travis.yml
```diff
@@ -8,15 +8,36 @@ sudo: false
...
-    - fish
+
+env:
+  global:
+    - PATH="$HOME/fish/bin:$PATH"
+    - OMF_REPO_URI="https://github.com/$TRAVIS_REPO_SLUG"
+    - OMF_REPO_BRANCH="$TRAVIS_BRANCH"
+
+before_install:
+  - tests/install_fish.sh
```
bpinto: What is the benefit of installing fish like this?

.travis.yml
```diff
@@ -8,15 +8,36 @@ sudo: false
...
+  - fish -c 'ln -sf $TRAVIS_BUILD_DIR/tests/fish_user_function $fish_function_path[1]'
     
 before_script: pwd; tree -h
 
-script: /bin/sh bin/install
+script:
+  - /bin/sh tests/travis_install.sh
+  - tests/print_prompt
```
bpinto: What is the goal here?

tests/travis_tests.fish
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env fish
+
+set -l return_code 0
+
+omf help
+omf query PATH
+omf install apt
+
+exit $return_code
```
bpinto: Always green tests?

bin/install
```diff
@@ -58,19 +58,16 @@ omf_install() {
...
   fi
 
-  pushd ${OMF_PATH} >/dev/null 2>&1
-
   local git_rev=$(git rev-parse HEAD) >/dev/null 2>&1
   local git_upstream=$(git config remote.upstream.url)
 
   if [ -z "${git_upstream}" ]; then
-    git remote add upstream ${git_uri}
+    git -C ${OMF_PATH} remote add upstream ${git_uri}
```
bpinto: How backwards compatible is it?

tests/fish_user_function/refresh.fish
```diff
@@ -0,0 +1,14 @@
+# SYNOPSIS
+#   refresh
+#
+# OVERVIEW
+#   Refresh (reload) the current fish session.
+
+function refresh -d "refresh the fish session"
+  if set -q CI
+    echo In travis !
```
bpinto: Perhaps a better self explanatory message could be used?

bpinto:
Great changes!!!

.travis.yml
jeremiejig: Well fish in the ubuntu repo is the version 1.23.1.
If we wan't to keep using the container-based of travis we have to build it ourselves.

.travis.yml
jeremiejig: About the` tests/print_prompt`, there wasn't any goal here.
Just wanted to see if the `print_prompt` behave correctly and kind of let it here.

.travis.yml
jeremiejig: This is to point the directory `~/.config/fish/fish_user_function` on the one present in the tests dir.
It allow us to override some function which may not work in the travis build. (like refresh).

bin/install
jeremiejig: Mmmh the first release of git with this option is [this commit](https://github.com/git/git/commit/44e1e4d67d5148c245db362cc48c3cc6c2ec82ca) which is the version 1.8.5, which dates back from 27 Nov. 2013.

If this isn't enough I can use the option `--git-dir` and `--work-tree` which are both present since [this commit](https://github.com/git/git/commit/892c41b98ae2e6baf3aa13901cb10db9ac67d2f3), which is the version 1.5.3 released the 07 June 2007.

tests/travis_tests.fish
jeremiejig: Because it was a draft, I didn't really pay attention to this.

I'll push another one with a small draft of what we may want here.

tests/fish_user_function/refresh.fish
```diff
@@ -0,0 +1,14 @@
+# SYNOPSIS
+#   refresh
+#
+# OVERVIEW
+#   Refresh (reload) the current fish session.
+
+function refresh -d "refresh the fish session"
+  if set -q CI
```
derekstavis: Do we need this check? Don't we ensure that `tests/fish_user_function` is loaded only on CI environment?

tests/fish_user_function/refresh.fish
derekstavis: Maybe `Shell refresh requested while in testing environment!`?

.travis.yml
derekstavis: I also don't like the directory `tests` containing scripts that aren't tests. Perhaps we could use a `tools` directory in the project root. What do you think, @bpinto?

.travis.yml
bpinto: Okay

.travis.yml
bpinto: So we were run oh the tests in an old version of fish until now? 

Tools does sound better

bin/install
bpinto: Safer to use --git-dir? I am not confident old osx  versions have a new git version.

tests/fish_user_function/refresh.fish
jeremiejig: Yeah, we don't need the check. That was a snippet from some testing.

bin/install
jeremiejig: Well I can implement some `omf::pushd` `omf::popd` to avoid modifying the fish state of the user.

bpinto: Easier to use git-dir?

derekstavis:
I think `tests/fish_user_functions` could be changed to `tests/function_overrides`, or something more meaningful.

jeremiejig:
I'd like to add some other thing to `tools/travis_tests.fish`.
Like some test counter and how many tests failed.

About the `fish_override_functions` I'm thinking of adding some color functions to prettify the tests logs :smile: .

derekstavis:
This is what I see as a good testing structure:

```
- tools
  * travis_install.sh
- tests
  - function_overrides
    * refresh.fish
  * it_installs_package.fish
  * it_removes_package.fish
  * it_updates.fish
```

Note the separation of concerns:
- `tools` directory include scripts needed by travis **before** running tests
- `tests/function_overrides` directory include function overrides for testing environment
- `tests/it_*.fish` are test files with a single function with the same file name, which are executed one by one
- `tests/test_runner.fish` is the main test file, where tests are called. Tests can be called automatically or added manually, we need to discuss it

We had `omf-spec` plugin some time ago, I think we could reuse concepts from it. I would like to hear @oh-my-fish/core to organize the test infrastructure in a flexible way, so after this merge we can start writing real tests, and maybe bringing `omf-spec` back to its duties.

jeremiejig:
Well I have to point out that there is 3 tests that have to be done separately: 

* The fresh installation of omf (already done)
* The removal of omf.
* The reinstallation of omf (with an already present config structure).

I think they are pretty important test.
Do we add this test like this:
```
- tools
   * travis_remove.fish
   * travis_reinstall.fish
```
or in the `tests` folder ?

The actual implementation of this two more test won't be the subject of this PR.

derekstavis:
When they get implemented I think that it's better to have this:

```
- tests
  * it_installs.fish
  * it_destroys.fish
  * it_installs_with_existing_config.fish
```

This way we separate fish install from framework install. What do you think about?

jeremiejig:
But how do we run them if the plugin isn't loaded anymore ?

What do you mean by fish install and framework install, do you mean respectively `bin/install` and `omf install` ?

derekstavis:
I see Oh My Fish test procedure like this:

- Travis fetches current build
- Travis install fish or uses cached
- **Tests begin**
- Travis test script runs `tests/test_runner.fish`
  * `it_installs.fish` install the framework using README steps
    + From now `omf` exists in the environment
  * `it_installs_package.fish` install some package and check if it is installs
  * `it_removes_package.fish` install that sane package and check if it is removed
  * `it_destroys.fish` destroy the framework (keeps configs)
  * `it_installs_with_existing_config.fish` install the framework with existing config files

jeremiejig:
Well I do like the basic structure.

I've just moved and renamed `travis_test` to `test_runner`.
I've just thougth of adding `osx` in `.travis.yml`. Do we want that ?

Although I think I need another PR for that.

.travis.yml
```diff
@@ -8,15 +8,35 @@ sudo: false
...
+
+before_install:
+  - tools/travis_install_fish.sh
+  - fish -c 'ln -sf $TRAVIS_BUILD_DIR/tests/functions_override $fish_function_path[1]'
     
 before_script: pwd; tree -h
 
-script: /bin/sh bin/install
+script:
+  - /bin/sh tools/travis_install.sh
```
bpinto: Can we rename this to Travis install omf? I'm always confused when I read the other file Travis install fish.

bpinto:
The discussion above is awesome but may we merge this and discuss it on another PR or issue? I'd like to ship what we have. Move fast and safe :)

bpinto:
Squash some commits and rebase?

bpinto:
:+1:

bpinto:
Ping @jeremiejig

jeremiejig:
Just rebased on master, but didn't really look for the squashing.

Do you want me to squash the rename of directories ? or other thing ?

derekstavis:
There are some squashes and edits you can do: For example, you add `print_prompt` then in commit 179886a you remove it. I'd also try to keep a logical history and group by file or changeset i.e: group all pushd/popd in a series, but split by file.

derekstavis:
`a more correct travis_tests` could be renamed to `tests: test commands help, query & install`

derekstavis:
`simplfying refresh from tests dir.` could be squashed with the first commit which creates the respective file.

derekstavis:
I think from commit a5b2c80 you can squash into previous, as this range just overlay more changes in files already added or changed.

pkg/omf/cli/omf.repo.pull.fish
```diff
@@ -1,42 +1,49 @@
 function omf.repo.pull
-  if test (command git config --get remote.upstream.url)
+  set -l repo_dir
```
bpinto: Let's do this in a future PR, but instead of having this whole block of code, we could just call `omf.repo.pull $OMF_PATH` where we are calling it without an argument. :D

jeremiejig:
Ok done with the rebase.

bpinto:
Super! @jeremiejig

derekstavis:
Yay! Thanks @jeremiejig <3

PR #53, opened 2015-09-17
add reference to package 'fzf'

gretel:
A command-line fuzzy finder written in Go


bpinto:
We can't control a plugins code, but we advise you to split your functions on separate files, this will allow fish to load faster and better support function updates/reloads.

The completion file is empty as some other functions and could be removed.

:+1: for the merge.

derekstavis:
:+1:

PR #55, opened 2015-09-19
Add generate-themes-doc.fish

Pyppe:
Wanted to browse thru all the available themes, but could not find a centralized place for this. This PR contains a `bin/generate-themes-doc.fish` which generates `docs/Themes.md` that can be used for previewing the themes.

See https://github.com/Pyppe/oh-my-fish/blob/master/docs/Themes.md

### DIFF
--- a/tools/generate-themes-doc.fish
+++ b/tools/generate-themes-doc.fish
@@ -1,0 +1,55 @@
+#!/usr/bin/env fish
+
+set -l project_dir (status -f|xargs dirname|xargs dirname)
+set -l theme_doc "$project_dir/docs/Themes.md"
+set temp_theme_contents (mktemp /tmp/fish.Themes.Content.XXXXX.md)
+set temp_theme_toc (mktemp /tmp/fish.Themes.TOC.XXXXX.md)
+
+function __write_theme_readme -a name raw_content readme
+  set -l escaped_raw_content (echo $raw_content | sed -r s"#[\/]#\\\/#gi")
+  curl -s "$raw_content/$readme" | \
+    sed -r 's/^#(.*)$/###\1/gi' | \
+    sed -r 's/^=+$//gi' | \
+    sed -r 's/^!\[(.*)\]\(([^h][^t][^t].*)\)$/![\1]('$escaped_raw_content'\/\2)/gi' \
+    >> $temp_theme_contents
+  echo "" >> $temp_theme_contents
+  echo "" >> $temp_theme_contents
+end
+
+function __find_readme -a raw_content
+  for readme in "README.md" "readme.md"
+    if curl -Ifs "$raw_content/$readme" > /dev/null
+      echo $readme
+      break
+    end
+  end
+end
+
+echo "Generating Themes documentation ..."
+echo "# Available themes" > $temp_theme_toc
+
+for theme in (command find $project_dir/db/themes/ -type f|sort)
+  set -l name (echo $theme|xargs basename)
+  set -l url (cat $theme)
+  set -l raw_content (echo $url|sed -r 's#https://github.com/([-.a-z0-9]+)/([-.a-z0-9]+)#https://raw.githubusercontent.com/\1/\2/master#gi')
+  set -l readme (__find_readme $raw_content)
+
+  echo "- [$name](#$name)" >> $temp_theme_toc
+  echo "# $name" >> $temp_theme_contents
+  if [ $readme ]
+    echo "Fetching readme for $name"
+    __write_theme_readme $name $raw_content $readme
+  else
+    echo "FAILED: No readme for $name" 1>&2
+    echo "See $url for details" >> $temp_theme_contents
+  end
+end
+
+echo "" >> $temp_theme_toc
+cat $temp_theme_contents >> $temp_theme_toc
+rm $theme_doc
+cat $temp_theme_toc >> $theme_doc
+echo "All done: $theme_doc"
+
+rm $temp_theme_contents
+rm $temp_theme_toc
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ List installed packages.
 
 #### `omf theme` _`<theme>`_
 
-Apply a theme. To list available themes type `omf theme`.
+Apply a theme. To list available themes type `omf theme`. You can also [preview available Themes](./docs/Themes.md) before installing.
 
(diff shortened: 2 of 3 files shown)


bpinto:
I love this!!! Is there a way to automate it though?

derekstavis:
Thanks!! I like having this docs!! By the way, the script could be improved by generating a table of contents on the top of the docs. What do you think about?

Pyppe:
Thanks for the feedback, guys! I improved the script, and added TOC.

Regarding automation; I don't have a silver bullet. Of course a custom script could be added that overrides the generated `docs/Themes.md` and pushes it to master if it was changed... And then this could be invoked e.g. for each Travis CI-build. But personally I wouldn't want this kind of automagic behavior for my project... :)

Hmm... after-thought: Travis could generate the `Themes.md` and send email if it has changed. (Although, I don't know how easy/possible this kind of automation is to implement in Travis).

Pyppe:
Regarding automation, I think this could be one solution https://github.com/Pyppe/oh-my-fish/commit/a898f84691dcf133c0783d33fb1ec0a55db6a3b7: It runs the `generate-themes-doc.fish` as a Travis-test, and the build fails if `Themes.md` has changed.

Travis seems to be having some issues (https://twitter.com/pyrylahti/status/645578947560873985); so I'm yet to verify that this actually works...

derekstavis:
Looks awesome with the TOC!! I was also wondering if we should include a "Back to top" link after every theme, as navigation gets hard on such big files.

derekstavis:
Regarding to the place where the script is stored, maybe we should add this category of scripts to a `tools` directory, @oh-my-fish/core?

bpinto:
I like using Travis for the automation! Very clever. Maybe tools is a good folder.

derekstavis:
So I propose the following:

    bin/generate-themes-docs.fish -> tools/generate-themes-docs.fish

As I believe that PR #52 will also produce some other scripts that fit into `tools` directory.

Pyppe:
* Changed the directory from `bin/` to `tools/`
* Updated `README.md` to mention `Themes.md`

IMO we could already merge this as is. I can add the Travis integration later once #52 gets merged (because it seems to do some of the same things I'd like to do).

tools/generate-themes-doc.fish
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/fish
+
+set -l project_dir (status -f|xargs dirname|xargs dirname)
+set -l theme_doc "$project_dir/docs/Themes.md"
+set temp_theme_contents (mktemp /tmp/fish.Themes.Content.XXXXX.md)
```
bpinto: Does mktemp work on Linux and OS X?

tools/generate-themes-doc.fish
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/fish
```
bpinto: Get fish binary using env?

tools/generate-themes-doc.fish
```diff
@@ -0,0 +1,55 @@
...
+
+set -l project_dir (status -f|xargs dirname|xargs dirname)
+set -l theme_doc "$project_dir/docs/Themes.md"
+set temp_theme_contents (mktemp /tmp/fish.Themes.Content.XXXXX.md)
+set temp_theme_toc (mktemp /tmp/fish.Themes.TOC.XXXXX.md)
+
+function write_theme_readme -a name raw_content readme
```
bpinto: Can we follow the guidelines for private functions?

tools/generate-themes-doc.fish
```diff
@@ -0,0 +1,55 @@
...
+      echo $readme
+      break
+    end
+  end
+end
+
+echo "Generating Themes documentation ..."
+echo "# Available themes" > $temp_theme_toc
+
+for t in (command find $project_dir/db/themes/ -type f|sort)
```
bpinto: Theme instead of t?

tools/generate-themes-doc.fish
```diff
@@ -0,0 +1,55 @@
...
+  set -l raw_content (echo $url|sed -r 's#https://github.com/([-.a-z0-9]+)/([-.a-z0-9]+)#https://raw.githubusercontent.com/\1/\2/master#gi')
+  set -l readme (find_readme $raw_content)
+
+  echo "- [$name](#$name)" >> $temp_theme_toc
+  echo "# $name" >> $temp_theme_contents
+  if [ $readme ]
+    echo "Fetching readme for $name"
+    write_theme_readme $name $raw_content $readme
+  else
+    echo "FAILED: No readme for $name"
```
bpinto: Output to stderr?

tools/generate-themes-doc.fish
Pyppe: IMO, yes.

Pyppe:
Made changes. Some questions, though:
* What do you mean by "Get fish binary using env?"
* Renamed scripts functions with `__` prefix? Is this enough / OK?

bpinto:
The prefix is perfect! Regarding env, this is what I mean: http://stackoverflow.com/questions/2429511/why-do-people-write-usr-bin-env-python-on-the-first-line-of-a-python-script

Pyppe:
Roger. Changed the shebang-line to `#!/usr/bin/env fish`

bpinto:
Really nice!!! Eager to see the automation using Travis and perhaps a similar feature for plugins?

:+1: for the merge. Would you mind squashing your commits into one?

derekstavis:
I've tested here, no issues. :+1: for the merge. Also, rebasing into current master would be great to keep the history linear :)

Pyppe:
Rebased and squashed. :shipit:

derekstavis:
Super!! Thanks @Pyppe!!

bpinto:
@Pyppe we are back. Whenever you have a chance to work on the travis script, submit a PR! :dancers:

Pyppe:
I've got the functionality here: https://github.com/Pyppe/oh-my-fish/commit/de464e56b1b401b5e9bc83f604ffe16b4d991244

However, I think we'd need some changes from the #52 PR in order to run these fish-based scripts within Travis. See https://travis-ci.org/Pyppe/oh-my-fish/builds/82388150

bpinto:
It has been merged! :dancers:

PR #62, opened 2015-09-26
Replace gitter with slack

bpinto:
We are using slackin (https://github.com/rauchg/slackin) to automatically accept people's invitation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,8 @@
 
 > The <a href="http://fishshell.com">Fishshell</a> Framework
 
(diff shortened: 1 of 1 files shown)


derekstavis:
Maybe this badge looks better?



image source:

https://img.shields.io/badge/slack-join%20the%20chat-38978D.svg?style=flat-square

derekstavis:
In the future we could add the possibility of tweaking badges to slackin

bpinto:
I'd rather have the count on the badge and work on the badge in the future.

The count works as a way to show there are people online in the community in case you need help.

derekstavis:
So let's go for it :+1:

derekstavis:
Ah, don't forget to update README links too!!

bpinto:
Which links?

PR #63, opened 2015-09-26
Travis integration for slack

bpinto:
Command: `travis encrypt "..." --add notifications.slack`


derekstavis:
Looks like travis worked with it, so :+1:

PR #64, opened 2015-09-27
Plugin dependency

bpinto:
@derekstavis could you confirm this works with `nvm`?

### DIFF
--- a/lib/autoload.fish
+++ b/lib/autoload.fish
@@ -1,15 +1,22 @@
 # SYNOPSIS
-#   autoload <path [path...]>
+#   autoload <path>...
 #
 # OVERVIEW
 #   Autoload a function or completion path. Add the specified list of
 #   directories to $fish_function_path. Any `completions` directories
 #   are correctly added to the $fish_complete_path.
+#
+#   Returns 0 if one of the paths exist.
+#   Returns 1 if all paths are missing.
 
 function autoload -d "autoload a function or completion path"
+  set -l path_exist "false"
+
   for path in $argv
+    set -l dest fish_function_path
+
     if test -d "$path"
-      set -l dest fish_function_path
+      set path_exist "true"
 
       if test (basename "$path") = completions
         set dest fish_complete_path
@@ -18,4 +25,6 @@ function autoload -d "autoload a function or completion path"
       contains "$path" $$dest; or set $dest "$path" $$dest
     end
   end
+
+  test $path_exist = "true"; and return 0; or return 1
 end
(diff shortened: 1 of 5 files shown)


derekstavis:
Awesome! I love it! Testing right now!

derekstavis:
It works amazingly well. While working on #8 we could also include automagic dependency install/loading too, as this PR just supports imperative dependencies, instead of a declarative ones.

bpinto:
True... Perhaps we should delay this PR until we add support for that? I think the way to use this function will be quite different.

bpinto:
I will work on some other things that I feel are necessary before I can dive on this. :)

derekstavis:
Nice! I think that we could wait to tackle dependencies in a row!

derekstavis:
Also, why do this PR addresses `omf destroy`?

bpinto:
You are too pedantic. :)

lib/autoload.fish
```diff
@@ -1,15 +1,22 @@
...
 
 function autoload -d "autoload a function or completion path"
+  set -l path_exist "false"
+
   for path in $argv
+    set -l dest fish_function_path
+
     if test -d "$path"
-      set -l dest fish_function_path
+      set path_exist "true"
```
derekstavis: You can just `set -l path_exist` and in the end `set -q path_exist; and return 0; or return 1`.

bpinto: Better!

lib/available.fish
```diff
@@ -4,6 +4,6 @@
 # OVERVIEW
 #   Check if a program is available.
 
-function available -a program -d "check if a program is available."
-  type "$program" ^/dev/null >&2
+function available -a name -d "Check if a program is available."
```
derekstavis: Leveraging this change, `Check if a function or program is available` describes the command better.

lib/require.fish
```diff
@@ -0,0 +1,33 @@
...
+#     - Emit its initialization event.
+#
+#   If the required plugin has already been loaded, does nothing.
+#   If the required plugin does not exist, return an error code and send a message to stderr.
+
+function require -a name
+  # Skip if plugin has already been loaded.
+  echo $fish_function_path | grep -q /pkg/$name; and return 0
+
+  set -l plugin_exist "false"
```
derekstavis: you can apply the same idea of `set -q` here.

derekstavis:
Great!! Cleaner code!! :star2:

bpinto:
Closing as this is not complete. We should work towards having a `bundle` file per plugin.

PR #65, opened 2015-09-27
Remove src attribute from spacer image in the README

nelsonjchen:
Spacer still works just as good.

This will make Chrome and its siblings not display a broken image placeholder
in the spacer.

Confirmed OK on Firefox and Safari.

screen shot 2015-09-26 at 7 52 58 pm


bpinto:
There are so many arrows that it took me 5 seconds to understand what you were trying to highlight! :joy: 

Thanks!

PR #68, opened 2015-09-27
Add tmux-zen plugin to db

coderstephen:
Adds my highly opinionated tmux plugin, [tmux-zen](https://github.com/coderstephen/tmux-zen), to the package database.


bpinto:
:+1: Will try this plugin shortly!

PR #71, opened 2015-09-27
Add test-generate-themes-doc.fish

Pyppe:
* Run with Travis CI
* Fail build if `docs/Themes.md` is no longer up-to-date

Here's Travis examples:
* [Build fails](https://travis-ci.org/Pyppe/oh-my-fish/builds/82439638)
* [Build succeeds](https://travis-ci.org/Pyppe/oh-my-fish/builds/82439947)

The iffy thing here is that whenever any theme README gets updated, the Travis-build will fail... Any thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ before_script: pwd; tree -h
 script:
   - /bin/sh tools/travis_install_omf.sh
   - tests/test_runner.fish
+  - tests/test-generate-themes-doc.fish
 
 after_script:
   - cd ~/.config/fish; tree -h; find . -type f | xargs cat
--- a/docs/Themes.md
+++ b/docs/Themes.md
@@ -34,6 +34,7 @@
 - [led](#led)
 - [mtahmed](#mtahmed)
 - [nai](#nai)
+- [nelsonjchen](#nelsonjchen)
 - [numist](#numist)
 - [ocean](#ocean)
 - [perryh](#perryh)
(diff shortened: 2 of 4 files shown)


derekstavis:
This is definitely something! I think we can make Travis deploy the generated docs. I have found this: http://awestruct.org/auto-deploy-to-github-pages/#automated-publishing-using-travis-ci

bpinto:
> The iffy thing here is that whenever any theme README gets updated, the Travis-build will fail... Any thoughts?

Perhaps we should only run this step if a change was done to `db` folder?

derekstavis:
>Perhaps we should only run this step if a change was done to `db` folder?

This will also make Travis don't exceed the API rate limit :)

Pyppe:
Added conditional execution by using `https://api.github.com/repos/$TRAVIS_REPO_SLUG/compare/$TRAVIS_COMMIT_RANGE`. Used Github API, because couldn't find list of changes files directly from Travis. See commit: https://github.com/oh-my-fish/oh-my-fish/commit/938fc5c810aab2dd56abf162cb7c655250397ee8?w=1

bpinto:
Awesome, looking good to me! :+1:

derekstavis:
Awesome @Pyppe, thanks again for your work. You mind squashing and rebasing to master?

Pyppe:
Rebased & squashed :shipit:

derekstavis:
Go for it @bpinto :dancer:

bpinto:
Done! Thanks @Pyppe

PR #74, opened 2015-09-28
Add kill-on-port pkg

vincentjames501:
Is there a good way of specifying transient dependencies? This pkg requires the getopts pkg for example...?

### DIFF
--- a/db/pkg/kill-on-port
+++ b/db/pkg/kill-on-port
@@ -1,0 +1,1 @@
+https://github.com/vincentjames501/fish-kill-on-port


vincentjames501:
Ah. +1 on #64 :+1:

derekstavis:
As `getopt` comes alphabetically before `kill-on-port` you have no issues. We are working in solving dependencies in a declarative way instead of imperative, which should also auto-install them. Although not very recommended, you can do what I've done on `foreign-env` package.

bpinto:
Very interesting plugin!

bobthecow:
> As getopt comes alphabetically before kill-on-port you have no issues. 

Is this guaranteed, or does it depend on having a filesystem that returns directories in alphabetical order?

derekstavis:
I think it depends more on how the shell expands the directory listing.

PR #79, opened 2015-09-29
Move $OMF_CONFIG variable to $OMF_PATH/init.fish

derekstavis:
This PR sets the default value of `OMF_CONFIG` variable in framework `init.fish` file. The variable can still be overridden by the user by setting it on `~/.config/fish/config.fish file`.

This is in preparation to rewriting install script in plain fish, which will ditch config template and stop replacing `config.fish`contents in favor of just appending Oh My Fish startup.

### DIFF
--- a/init.fish
+++ b/init.fish
@@ -26,6 +26,12 @@ else
   set -gx ORIGINAL_PATH $PATH
 end
 
+# Set OMF_CONFIG if not set.
+if not set -q OMF_CONFIG
+  set -q XDG_CONFIG_HOME; or set -l XDG_CONFIG_HOME "$HOME/.config"
+  set -gx OMF_CONFIG "$XDG_CONFIG_HOME/omf"
+end
+
 # Save the head of function path and autoload core functions
 set -l user_function_path $fish_function_path[1]
 set fish_function_path[1] $OMF_PATH/lib
--- a/bin/install
+++ b/bin/install
@@ -89,8 +89,7 @@ omf_install() {
   touch ${fish_config_file} >/dev/null 2>&1
   test ! -w ${fish_config_file} && die "Fish configuration file is not writable, aborting."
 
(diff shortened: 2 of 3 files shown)


bpinto:
:+1: good to me. I'll add more documentation regarding `OMF_CONFIG`.

ping @bobthecow

init.fish
```diff
@@ -26,6 +26,11 @@ else
   set -gx ORIGINAL_PATH $PATH
 end
 
+# Set OMF_CONFIG if not overriden.
+if not set -q OMF_CONFIG
+  set -gx OMF_CONFIG "$HOME/.config/omf"
```
bobthecow: nope. we need to respect XDG, not just support its default value.

we need to replicate this (from the current install script):

```
 test -z ${XDG_CONFIG_HOME+_}  && XDG_CONFIG_HOME="${HOME}/.config"
 test -z ${OMF_CONFIG+_}       && OMF_CONFIG="${XDG_CONFIG_HOME}/omf"
```

derekstavis: True story, I overlooked it!! Thanks!!

derekstavis:
Done @bobthecow!

derekstavis:
Do we need to `set -gx OMF_CONFIG`?

bpinto:
We should need `-gx`.

I'm not familiar with XDG again... The README was too confusing (or the page was too ugly and I skipped the important parts :D). But should it be used only during install or on every load?

derekstavis:
Well, according to what I usually it should be used everytime. If one reinstall the system and the user path ($HOME) is migrated, XDG envs should be already set the accordingly.

bobthecow:
@bpinto Yeah, it should happen every time. A user should be able to toggle between multiple config directories by changing the environment variable. 

@derekstavis Thanks!

bobthecow:
:+1:

PR #83, opened 2015-09-30
Faster loading

bpinto:
- Initial work for supporting plugin dependencies.
- Source `init.fish` over `$package_name.fish`.

### DIFF
--- a/init.fish
+++ b/init.fish
@@ -47,8 +47,12 @@ for path in $OMF_PATH/lib $OMF_PATH/lib/git {$OMF_PATH,$OMF_CONFIG}/pkg/* $theme
   contains -- (basename $path) $OMF_IGNORE; and continue
 
   autoload $path $path/completions
-  source $path/(basename $path).fish ^/dev/null
-    and emit init_(basename $path) $path
+
+  if source $path/init.fish ^/dev/null
+  else
+    source $path/(basename $path).fish ^/dev/null;
+    #and echo "Plugin '"(basename $path)"' has a deprecated structure. Run `omf update`."
+  end; and emit init_(basename $path) $path
 end
 
 autoload $OMF_CONFIG/functions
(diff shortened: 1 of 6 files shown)


init.fish
```diff
@@ -41,16 +41,20 @@ source $OMF_CONFIG/before.init.fish ^/dev/null
...
 
-  autoload $path $path/completions
-  source $path/(basename $path).fish ^/dev/null
-    and emit init_(basename $path) $path
+  contains -- $name $OMF_IGNORE; and continue
+  require $name
 end
 
+# Autoload theme
+autoload {$OMF_PATH,$OMF_CONFIG}/themes/(cat $OMF_CONFIG/theme)
```
bpinto: feature

The idea is to support overwriting specific functions on your theme/plugin. So if you correctly split functions on individual files, one have the opportunity to change a specific part of it.

bobthecow:
```
function autoload_one -d "autoload the first directory which exists in the list of directories"
  for path in $argv
    autoload $path; and return
  end
end
```

then we could

```
autoload_one {$OMF_CONFIG,$OMF_PATH}/themes/(cat $OMF_CONFIG/theme)
```

?

(note that the OMF_PATH/OMF_CONFIG are swapped, so the local one wins if two have the same name)

bobthecow: yeah. kind of makes sense. also makes things potentially crazy :)

bpinto: That's true, IMHO the benefits of allowing this are bigger. We have supported this for a long time (not sure in years) and I don't remember we having a problem with it.

pkg/omf/init.fish
```diff
@@ -0,0 +1,3 @@
+function init -a path --on-event init_omf
```
bobthecow: i'm not a fan of how our plugins all leak `init` functions :(

bobthecow:
:+1:

pkg/omf/init.fish
bpinto: Yeah me neither... this kinda defeats the purpose of having the `init_omf` event. BUT, I couldn't think of a better and easier to use idea yet...

init.fish
bobthecow: :+1:

PR #84, opened 2015-09-30
Stash does not support --git-dir

bpinto:
Stash is a shell script that does not support `--git-dir`, not even on
latest versions. We need to use `git -C` which was only added on recent
versions of `git`.

`omf update` will fail in case omf need needs to run `git stash` and
git is not compatible.


derekstavis:
Shouldn't we check git version? Maybe adding this check on install script?

bpinto:
I think we are fine with git version. `stash` is currently broken and no reports so far. It's not going to be a very used feature, and preventing people that don't have a new version of `git` to use `omf` is worse if it's only required here.

bobthecow:
:+1:

PR #88, opened 2015-10-01
Save history before replacing shell process

derekstavis:
Fixes an issue with history between shell reloads via refresh function call.
As `exec` replaces the current process in a non-gently fashion, the
persistent history could be out of sync with the in-memory history.

Calling `history --save` before `exec` should force the shell to persist the
history before replacing the current process.

### DIFF
--- a/lib/refresh.fish
+++ b/lib/refresh.fish
@@ -4,6 +4,9 @@
 # OVERVIEW
 #   Refresh (reload) the current fish session.
 
-function refresh -d "refresh the fish session"
-  set -q CI; or exec fish < /dev/tty
+function refresh -d "Refresh fish session by replacing current process"
+  if not set -q CI
+    history --save
+    exec fish < /dev/tty
+  end
 end


bobthecow:
i just think we need to stop replacing the shell. it makes everything awful and does (almost) no good.

derekstavis:
yes, but while we don't solve the major issue we have to fix this specific history issue...

bpinto:
I agree with both, as a temporary solution, this seems good.
:+1:

bobthecow:
sure, :+1: until we fix the underlying problem.

derekstavis:
Agreed :+1:

lib/refresh.fish
```diff
@@ -4,6 +4,9 @@
 # OVERVIEW
 #   Refresh (reload) the current fish session.
```
bpinto: Refresh or reload? This comment is so weird... if `reload` is easier shouldn't we just call this function `reload`? :trollface:

derekstavis: Sure! `reload` makes much more sense!

PR #89, opened 2015-10-02
Update pkg urls and add weather plugin

coderstephen:


bpinto:
:+1: Nice plugins!

bobthecow:
:+1:

PR #94, opened 2015-10-03
Only emit uninstall_* event on `omf destroy`

bpinto:
When calling `omf destroy`, the `bundle` file is completely erased
because we are calling `remove_package` on each installed package. We
don't want to erase that file, we just want to emit the event and later
on remove the code, which is done with `rm -rf "$OMF_PATH"`.


derekstavis:
:+1:

PR #97, opened 2015-10-03
Fishy packages

bpinto:


pkg/omf/cli/omf.remove_package.fish
```diff
@@ -15,7 +15,7 @@ function omf.remove_package
   for path in {$OMF_PATH,$OMF_CONFIG}/{pkg}/$pkg
     not test -d $path; and continue
 
-    emit uninstall_$pkg
+    source $path/uninstall.fish; and emit uninstall_$pkg
```
bobthecow: for backwards compatibility for now, we should emit the event even if we can't source the uninstall file. drop it down to the next line with a TODO to change it to this way after a while?

bobthecow:
:+1: except the backwards compatibility break.

pkg/omf/cli/omf.remove_package.fish
bpinto: Yeah... but we have no plugins that have a `uninstall` function?

bobthecow: oh, i don't know. i haven't checked.

PR #99, opened 2015-10-04
Implement `autoload -e` option to remove from autoload paths

derekstavis:
Address the use case of autoload path removal, which allows themes
and packages definitions to be unloaded from running fish session.

This paves the way for allowing us to reload Oh My Fish without 
`exec`ing the shell, and thus overcoming the known issues with it.

### DIFF
--- a/lib/autoload.fish
+++ b/lib/autoload.fish
@@ -1,28 +1,64 @@
 # SYNOPSIS
 #   autoload <path>...
+#   autoload -e <path>...
 #
 # OVERVIEW
-#   Autoload a function or completion path. Add the specified list of
-#   directories to $fish_function_path. Any `completions` directories
-#   are correctly added to the $fish_complete_path.
+#   Manipulate autoloading path components.
 #
-#   Returns 0 if one of the paths exist.
-#   Returns != 0 if all paths are missing.
+#   If called without options, the paths passed as arguments are added to
+#   $fish_function_path. All paths ending with `completions` are correctly
+#   added to $fish_complete_path. Returns 0 if one or more paths exist. If all
+#   paths are missing, returns != 0.
+#
+#   When called with -e, the paths passed as arguments are removed from
+#   $fish_function_path. All arguments ending with `completions` are correctly
+#   removed from $fish_complete_path. Returns 0 if one or more paths erased. If
+#   no paths were erased, returns != 0.
 
-function autoload -d "autoload a function or completion path"
-  for path in $argv
+function autoload -d "Manipulate autoloading path components"
+  set -l paths $argv
+
+  switch "$argv[1]"
+  case '-e' '--erase'
+    set erase
+
+    if test (count $argv) -ge 2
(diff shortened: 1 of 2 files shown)


lib/autoload.fish
```diff
@@ -10,18 +10,46 @@
...
+
+  switch "$argv[1]"
+  case '-e' '--erase'
+    set erase
+
+    if test (count $argv) -ge 2
+      # Unshift path to remove option
+      set paths $argv[2..-1]
+    else
+      # Option called without arguments
```
bpinto: Missing `echo ... 1>&2`

lib/autoload.fish
```diff
@@ -1,5 +1,5 @@
 # SYNOPSIS
-#   autoload <path>...
+#   autoload [-e] <path>...
```
bpinto: Better explain `-e` usage? And include `--erase` if we are going to support it.

lib/autoload.fish
```diff
@@ -10,18 +10,46 @@
...
+      # Unshift path to remove option
+      set paths $argv[2..-1]
+    else
+      # Option called without arguments
+      return 1
+    end
+  end
+
+  for path in $paths
+
```
bpinto: Using three lines of code for a `continue`: one blank line and another two for code.

As this is not the main feature, I'd convert the three lines into `not test -d $path`; and continue`.

derekstavis: Refresh your page :dancers:

bpinto: Nothing changed. :)

derekstavis: Aww, I see your point. Single lineness :P

lib/autoload.fish
```diff
@@ -10,18 +10,46 @@
...
+      set -e function_path[$index]
+    else
+      contains -- "$path" $function_path; and continue
+      set function_path "$path" $function_path
     end
+
+    set $dest $function_path
   end
 
   set -q path_exist
```
bpinto: I don't believe this function return is correct for `autoload -e`. Might want to document it too.

lib/autoload.fish
```diff
@@ -10,18 +10,46 @@
...
 function autoload -d "autoload a function or completion path"
-  for path in $argv
+  set -l paths $argv
+
+  switch "$argv[1]"
+  case '-e' '--erase'
+    set erase
+
+    if test (count $argv) -ge 2
+      # Unshift path to remove option
```
bpinto: Unnecessary? It's a common thing to do.

derekstavis:
Fixed some issues pointed:

* Better documented `autoload -e` on function header
* Made function return code according to documentation
* Single-lineness fun with `continue`

derekstavis:
Please also give a look at this one @bobthecow

bpinto:
:+1: This is awesome!

bobthecow:
:+1:

PR #102, opened 2015-10-04
Rework omf theme code: Integrate with fish_config, ditch refresh calls

derekstavis:
**Depends upon #99**

The activation of a theme is now done in two phases of four steps:

+ Theme validation
  - If not installed, install target theme
  - Locate the current and target theme in `$OMF_CONFIG` or `$OMF_PATH`
  - Check existence or create fish user function path
  - Check for conflicting `fish_prompt`, abort if found
+ Theme activation
  - Remove current theme from autoloading paths
  - Add target theme to autoloading path
  - Persist selected theme to `$OMF_CONFIG/theme`
  - Link target theme's `fish_prompt` to user's

This approach brings two major advantages:

- Theme prompt now shows accordingly in `fish_config` web interface
- Faster theme changes, reducing it from seconds to some milis
- Avoids calling refresh, which makes users much more happy

Results:

* A faster theme changing experiece
Screencast

* Integration with fish_config
Fish Config Screenshot

### DIFF
--- a/pkg/omf/cli/omf.theme.fish
+++ b/pkg/omf/cli/omf.theme.fish
@@ -1,6 +1,31 @@
-function omf.theme
-  if not test -e $OMF_CONFIG/themes/$argv[1]
-    omf.install --theme $argv[1]
+function omf.theme -a target_theme
+  set -l current_theme (cat $OMF_CONFIG/theme)
+  test "$target_theme" = "$current_theme"; and return 0
+
+  set -l prompt_filename "fish_prompt.fish"
+  set -l user_functions_path (omf.xdg.config_home)/fish/functions
+
+  mkdir -p "$user_functions_path"
+
+  if not omf.check.fish_prompt
+    echo (omf::err)"Conflicting prompt setting."(omf::off)
+    echo "Run "(omf::em)"omf doctor"(omf::off)" and fix issues before continuing."
+    return 1
   end
-  echo "$argv[1]" > $OMF_CONFIG/theme
+
+  # Replace autoload paths of current theme with the target one
+  autoload -e {$OMF_CONFIG,$OMF_PATH}/themes/$current_theme
+  autoload {$OMF_CONFIG,$OMF_PATH}/themes/$target_theme
+
+  # Find target theme's fish_prompt and link to user function path
+  for path in {$OMF_CONFIG,$OMF_PATH}/themes/$target_theme/$prompt_filename
+    if test -e $path
+      ln -sf $path $user_functions_path/$prompt_filename; and break
+    end
+  end
+
+  # Persist the changes
+  echo "$target_theme" > "$OMF_CONFIG/theme"
+
+  return 0
 end
--- a/pkg/omf/omf.fish
+++ b/pkg/omf/omf.fish
@@ -120,10 +120,11 @@ function omf -d "Oh My Fish"
 
         omf.list_themes | column | sed -E "s/$regex/"(omf::em)"\1"(omf::off)"/"
         omf::off
-
       else if test (count $argv) -eq 2
+        if not contains -- $argv[2] (omf.list_installed_themes)
+          omf.install --theme $argv[2]; or return 1
+        end
         omf.theme $argv[2]
-        refresh
       else
         echo (omf::err)"Invalid number of arguments"(omf::off) 1^&2
         echo "Usage: $_ "(omf::em)"$argv[1]"(omf::off)" [<theme name>]" 1^&2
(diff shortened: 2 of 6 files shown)


pkg/omf/cli/omf.theme.fish
```diff
@@ -1,6 +1,64 @@
...
+    if not omf.install --theme "$target_theme"
+      return 1
+    end
   end
-  echo "$argv[1]" > $OMF_CONFIG/theme
+
+  set -l current_theme_path
+  set -l target_theme_path
+
+  for path in {$OMF_CONFIG,$OMF_PATH}/themes
```
bobthecow: don't we want to reverse these, so `$OMF_CONFIG` wins over `$OMF_PATH` if they both have the same theme name?

bobthecow:
:+1: to the idea, this is a huge improvement.

Do we need to do the symlink thing with right prompt and title and such, or does the `autoload -e` handle that? What happens if you switch from a theme with a right prompt to a theme without one?

bpinto:
#99 has been merged. Rebase?

pkg/omf/cli/omf.theme.fish
derekstavis: isn't this the case at this line?

derekstavis:
@bobthecow this uses `autoload -e` to remove previous theme path and then `autoload` the new theme directory. In this process fish already replaces the prompt when reloading functions. The symlink here is to keep in sync with `fish_config`.

pkg/omf/cli/omf.theme.fish
bobthecow: i don't think so. it's checking `$OMF_CONFIG/themes` first, then `$OMF_PATH/themes`. If the theme is found in `$OMF_CONFIG`, it's set as `$target_theme_path`, but if that same theme name is found in `$OMF_PATH`, it will be overwritten before leaving the for loop.

bobthecow:
:+1: cool. tested it and it works well. my only remaining concern is that *it seems like* the `$OMF_CONFIG` and `$OMF_PATH` precedence is backward. but i'm also sleepy and could be completely mistaken, so someone should confirm this :)

pkg/omf/cli/omf.theme.fish
derekstavis: Yes, you are right! I was sleepy in this case, haha. I will fix it. Thanks for pointing!

derekstavis:
You were right all the time @bobthecow. Now I fixed the load order so we give priority to `$OMF_CONFIG` over ` $OMF_PATH`.

pkg/omf/omf.fish
```diff
@@ -120,10 +120,8 @@ function omf -d "Oh My Fish"
 
         omf.list_themes | column | sed -E "s/$regex/"(omf::em)"\1"(omf::off)"/"
         omf::off
-
       else if test (count $argv) -eq 2
         omf.theme $argv[2]
-        refresh
```
bpinto: This is beautiful.

pkg/omf/cli/omf.theme.fish
```diff
@@ -1,6 +1,64 @@
-function omf.theme
-  if not test -e $OMF_CONFIG/themes/$argv[1]
-    omf.install --theme $argv[1]
+function omf.theme -a target_theme
+  set -l current_theme (cat $OMF_CONFIG/theme)
+
+  test "$target_theme" = "$current_theme"; and begin; echo Already at; return 0; end
```
bpinto: Weird output: "Already at"

bpinto: Unnecessary `return 0`?

derekstavis: Oh. This was debug output. What we want here?

bpinto: Do we need anything? not sure

derekstavis: We can just `return 0` here to ensure prompt stays clean.

pkg/omf/cli/omf.theme.fish
```diff
@@ -1,6 +1,64 @@
...
+
+  if not contains -- "$target_theme" (omf.list_installed_themes)
+    if not omf.install --theme "$target_theme"
```
bpinto: I believe this is not part of the theme method. The CLI should know that we need to install the theme and then call omf theme.

pkg/omf/omf.fish
derekstavis: I would say pleasureful from developer and user point of view :dancers:

pkg/omf/cli/omf.theme.fish
derekstavis: Ok, I will move current theme check out of this function direct to the cli.

derekstavis:
Rebased to master.

derekstavis:
I have reworked and splitted some parts of code, also integrated with `omf doctor`, which was missing. Can you give another try @bobthecow @bpinto?

bpinto:
This is a great PR! Awesome work here @derekstavis

PR #105, opened 2015-10-05
Fix omf.destroy: -regextype doesn't exist on find BSD.

jeremiejig:
fish_config wasn't properly initialised.
GNU find use emacs regex by default. force using posix-basic (default of BSD find)


derekstavis:
@jeremiejig thanks for your attention to this, but this specific issue is solved by #102 which should be merged very soon :)

bpinto:
Could you confirm master is working @jeremiejig ?

Thanks for the PR!

jeremiejig:
Yep no problem for me with this master :smile: .
I only did some quick test, but seems like omf destroy, and omf install work as intended.

PR #112, opened 2015-10-08
update some cli completions: remove query and version. add doctor

nwykes:
I noticed the doctor command had been added put didn't see it in the completions.  Also remove the no longer valid query and version.


derekstavis:
Well done! Thanks @nwykes!! :+1:

bpinto:
Nice!!!

PR #113, opened 2015-10-12
refact(pkg/thefuck): use a better, oh-my-fishy thefuck plugin

scorphus:
We're good to go using our very own thefuck plugin. OMF FTW! Please review, and comment, these are much welcome.

### DIFF
--- a/db/pkg/thefuck
+++ b/db/pkg/thefuck
@@ -1,1 +1,1 @@
-https://github.com/wa/pkg-thefuck
+https://github.com/oh-my-fish/plugin-thefuck


bucaran:
Would be better to send a PR to the original package with your changes instead of creating a new package. 

The maintainer there is @gretel

gretel:
i don't think having duplicate packages is a good idea, too. regards

derekstavis:
:+1: for the merge.

bpinto:
Unfortunately the plugin @scorphus submitted is already using the [new architecture](https://github.com/oh-my-fish/oh-my-fish/issues/8).

scorphus:
@bpinto said:
> The plugin @scorphus submitted is a new plugin that uses the new architecture, which is required by the revamped omf.

Yeah, there's this incompatibility issue...

@bucaran said:
> Maybe better to send a PR to the original package with your changes instead of creating a new package?
> 
> @gretel

... so I opened wa/pkg-thefuck#1 as well.

@bpinto said:
> Thanks for the contribution @scorphus!

Always a pleasure :+1:

bpinto:
:dancers: :tada:

bucaran:
@bpinto Oh, so you are not sourcing `<package name>/<package name>.fish` anymore?

bpinto:
We won't source that file anymore, it's only sourcing for backwards compatibility.

PR #115, opened 2015-10-12
Merge all omf list functions

bpinto:

PR #118, opened 2015-10-13
Fixes a regression in update process added in #115

derekstavis:
As #115 migrated list functions, after updating Oh My Fish code
the omf.list_installed_packages gets removed, but is still being
used by the update code.

This commit fixes it by providing a wrapper function.


bobthecow:
Let's add a TODO with a removal date? Since the regression is mostly an awkward experience, not a complete failure, I'd be okay with that date being ~3 months out or so.

derekstavis:
`TODO: Remove it after 2015, December 13`?

bobthecow:
sure.

derekstavis:
done.

Issue #119, opened 2015-10-13
omf update error: omf package already installed.


gperaza:
When attempting to update omf with "omf update" I get the following error.
Error: omf package already installed.

derekstavis:
It's a known bug introduced by #115. There's a upcoming PR which will fix it. But thanks for reporting. We will keep this issue updated with the progress!

bpinto:
Fix has been merged.

PR #121, opened 2015-10-14
Overhaul install and update code

derekstavis:
This series breaks install code monolith by splitting
the update code out of it and also by rewriting install
code.

Also fixes some known issues:

- #119 which tries to update local packages and even
omf package
- Installing plugins form URLs ended in `.git` now
works correctly
- When using `omf install <package>`, don't refresh 
the shell if nothing was installed

This series also changes the behavior of `omf theme`,
which now don't install or update themes when used,
rather displaying a message guiding the user to use
`omf install`.


derekstavis:
:+1:

scorphus:
:+1:

PR #122, opened 2015-10-14
pkg/omf/init.fish: source `omf.fish` as part of omf's init

scorphus:
Since ceb31c1, functions defined by omf package – namely `omf::em`, `omf::dim`, `omf::err`, `omf::off` and `omf` – are not being sourced.

### DIFF
--- a/pkg/omf/init.fish
+++ b/pkg/omf/init.fish
@@ -1,3 +1,19 @@
 function init -a path --on-event init_omf
   autoload $path/cli $path/util
 end
+
+function omf::em
+  set_color $fish_color_match ^/dev/null; or set_color cyan
+end
+
+function omf::dim
+  set_color $fish_color_autosuggestion ^/dev/null; or set_color 555
+end
+
+function omf::err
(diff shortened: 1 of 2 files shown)


derekstavis:
This would make `omf::` functions available in the global scope. I'm not sure if we want this.

scorphus:
Yet, let me take a different approach. For I like those handy functions and use them in in my plugins... but I got it wrong.

derekstavis:
The idea is to switch to `plugin-technicolor` ASAP. Maybe it's time to open a PR?

bpinto:
I believe technicolor is a better solution, how close is it?

derekstavis:
So close that I should open a PR dropping it into this repository!

scorphus:
Go for it, @derekstavis! As soon as `technicolor` is in, I no longer need these `omf::` fellas

derekstavis:
I should open a PR this weekend. The basic DSL is done, I just need to sort out some details and document the available functions and how they can be composed.

PR #124, opened 2015-10-15
require: Cleanup previously sourced init functions

derekstavis:
`require` sources plugin's `init.fish` files, which are supposed to
declare an `init` function. Although we source the file, we aren't
cleaning up the declared function. This PR fixes this.

### DIFF
--- a/lib/require.fish
+++ b/lib/require.fish
@@ -23,4 +23,8 @@ function require -a name
         and emit init_$name $path
     end
   end
+
+  functions -e init  # Cleanup previously sourced function
+
+  return 0
 end


bobthecow:
i still don't like that the function is an un-namespaced "init". it's just asking for problems :(

derekstavis:
We should definitely use `<packagename>.init` in the future.

bpinto:
@derekstavis We used to have `init.load` files, and the problem with that is that the extension is not `.fish`.

I don't know see much problems with having it named `init.fish`. It's weird as the function created inside is not called `init` and the events can only be used one single time as the next plugin will override the function (so what's the point of using events?). But, I think it works just fine.

Issue #123, opened 2015-10-15
basename: illegal option


th3m4ri0:
Hey guys, trying to upgrade omf. I get a weird error breaking everything :'(

On Mac OS X 10.10.5 :
```bash
rm -rf ~/.oh-my-fish/
rm ~/.config/fish/config.fish
curl -L github.com/oh-my-fish/oh-my-fish/raw/master/bin/install > install
fish install
Installing Oh My Fish to /Users/ct/.local/share/omf...
Cloning master from https://github.com/oh-my-fish/oh-my-fish.git...
Adding startup code to fish config file...
Building Oh My Fish configuration...
basename: illegal option -- -
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
Error:  package already installed.
Installation successful!
```

th3m4ri0:
On Ubuntu 14.04 : 
```
ct@nuke ~> fish install
Installing Oh My Fish to /home/ct/.local/share/omf...
Cloning master from https://github.com/oh-my-fish/oh-my-fish.git...
Adding startup code to fish config file...
Building Oh My Fish configuration...
basename: unrecognized option '--theme'
Try 'basename --help' for more information.
Error:  package already installed.
Installation successful!
ct@nuke ~>
```

PR #125, opened 2015-10-15
Rewrite bundle install to conform with latest `omf.install` interface

derekstavis:
Rewrites the code, as now `omf.install` is type
agnostic, and any -- preceded parameter will
be interpreted as package name.

In case of #123 it was breaking on a 
`basename` call just for the fun of
debugging :trollface:.

### DIFF
--- a/pkg/omf/cli/omf.bundle.install.fish
+++ b/pkg/omf/cli/omf.bundle.install.fish
@@ -2,30 +2,28 @@ function omf.bundle.install
   set bundle $OMF_CONFIG/bundle
 
   if test -f $bundle
-    set packages (omf.packages.list --installed --plugin)
-    set themes (omf.packages.list --installed --theme)
+    set packages (omf.packages.list --installed)
     set bundle_contents (cat $bundle | sort -u)
 
     for record in $bundle_contents
-      set type (echo $record | cut -d' ' -f1)
-      set name_or_url (echo $record | cut -d' ' -f2-)
+      test -n "$record"; or continue
+
+      set type (echo $record | cut -s -d' ' -f1 | sed 's/ //g')
+      contains $type theme package; or continue
+
+      set name_or_url (echo $record | cut -s -d' ' -f2- | sed 's/ //g')
+      test -n "$name_or_url"; or continue
+
       set name (basename $name_or_url | sed 's/\.git//;s/^pkg-//;s/^plugin-//;s/^theme-//')
 
-      switch $type
-      case "package"
-        if not contains $name $packages
-          omf.install --pkg $name_or_url
-        end
-
-      case "theme"
-        if not contains $name $themes
-          omf.install --theme $name_or_url
-        end
+      if not contains $name $packages
+        omf.install $name_or_url;
+          and set installed
       end
     end
 
     sort -u $bundle -o $bundle
   end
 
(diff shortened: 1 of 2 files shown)


th3m4ri0:
Thanks for all the work @derekstavis. Working great on my side.

bobthecow:
I don't think we should be deleting lines from bundle we don't recognize. If…

1. We add a new type of line to it in the future (say, `colorscheme`)
2. A user syncs their dotfiles (via Dropbox or whatever)
3. … from a machine with this new omf version to one with the version we're about to land
4. And the user runs `omf i` before `omf u`

… we'll delete any `colorscheme` row from their bundle, just because the stale version of `omf` didn't recognize it. Even running `omf update` afterword won't bring it back. And better yet, because our hypothetical user was syncing via Dropbox, this change will be propagated to all their other boxes and they'll lose this new row type on all boxes! Sweet!

I'd prefer warning when we `continue`, dropping all the `$filtered_bundle` business, and changing line 27 back to `sort -u`

pkg/omf/cli/omf.bundle.install.fish
```diff
@@ -7,24 +7,24 @@ function omf.bundle.install
...
+      set name_or_url (echo $record | cut -s -d' ' -f2- | sed 's/ //g')
+      test -n "$name_or_url"; or continue
 
-      case "theme"
-        if not contains $name $themes
-          omf.install --theme $name_or_url
-        end
+      set name (basename $name_or_url | sed 's/\.git//;s/^pkg-//;s/^plugin-//;s/^theme-//')
+
+      if not contains $name $packages $themes
```
bobthecow: i know it's outside this diff, but we should combine the two `$packages` and `$themes` into one `$packages` …

above:

```fish
set packages (omf.packages.list --installed)
```

here:

```fish
if not contains $name $packages
```

pkg/omf/cli/omf.bundle.install.fish
```diff
@@ -7,24 +7,24 @@ function omf.bundle.install
...
+        omf.install $name_or_url
```
bobthecow: since `omf.install` has a useful status code, we could collect it here and change the `return 0` below so that `omf install` doesn't have to refresh unless it's necessary.

derekstavis: This was the intention, I just forgot to add this :P

pkg/omf/cli/omf.bundle.install.fish
derekstavis: I will change this, makes a lot of sense and makes things faster

derekstavis:
We're back from filtering the bundle. Shame on me, it was very outside of this PR scope. Thanks for poiting. When you have time, please give another try at this @bobthecow @bpinto

pkg/omf/omf.fish
```diff
@@ -76,17 +76,18 @@ function omf -d "Oh My Fish"
...
-            set refresh
-          end
+          omf.install $package;
+            and set refresh
         end
-
-        set -q refresh; and refresh
       end
 
+      set -q refresh; and refresh
```
bpinto: Weird to have the variable name the same as the function. Rename the variable?

PR #127, opened 2015-10-18
omf/templates: Remove Travis from scaffolding

derekstavis:
This PR remove Travis from package scaffolding template.

Testing should be opt-in, not opt-out.


bpinto:
> Testing should be opt-in, not opt-out.

Not sure if I agree, but we are not using it anyway. :D

PR #135, opened 2015-10-20
add https:// to curl calls in README.md

jrobeson:
My version of curl (at least) doesn't assume https:// even thought HSTS is probably set.


derekstavis:
You mind also adding it to `bin/install` script? Thanks!

jrobeson:
superseded by #137.  This one was done in the webui, which works fine for 1 file, but i did #137 on a real branch

Issue #131, opened 2015-10-18
Replace image logos by an optimised SVG version


derekstavis:
We are currently using a PNG-based logo which weights 18.80k. I have redrawn our logo in fully vectorised SVG, which brings benefits in scalability -- we need just one file and can abuse on dimensions -- and in load time -- I optimised it down to just 0.7k.

Please give a look below and tell what you guys think about:

![SVG Optimised Logo](https://dl.dropboxusercontent.com/u/56336/omf/omf-logo-optimised.svg)

scorphus:
I think it's bit too *roundy*. The original is more angular:

![Oh-My-Fish](https://avatars3.githubusercontent.com/u/11728505)

scorphus:
Remarkable job btw, @derekstavis, :+1:

derekstavis:
It's a downside of optimising the SVG file. If we remove floating point things get roundy :stuck_out_tongue:

scorphus:
I think there still going to be two points in each those curves. Are you using Inkscape? Or what?

derekstavis:
But yeah, in the tail things got more roundy, but also more concise, as rounding was applied homogeneously through the polygons.

derekstavis:
Adobe Illustrator, always!

![screen shot 2015-10-18 at 7 24 51 pm](https://cloud.githubusercontent.com/assets/1611639/10566729/edb0aabe-75cd-11e5-8ceb-dfc3a21070b3.png)

scorphus:
Argh... Let's keep it all open source, mate! Inkscape FTW!

derekstavis:
I'm familiar with Inkscape, but as I've worked with Illustrator for a long time I'm more used to its commands and "its way". I have redrawn the logo in about 8 steps.

scorphus:
![clip](https://cloud.githubusercontent.com/assets/2112697/10566753/8ecf7fc4-75ce-11e5-818d-1abb5b74688d.png)

derekstavis:
The original version is disrespecting the rounding harmony:

![screen shot 2015-10-18 at 7 31 09 pm](https://cloud.githubusercontent.com/assets/1611639/10566760/db3ea718-75ce-11e5-8ada-e7ef437ea535.png)

Left side is very *roundy*, while right side is almost straight!!

scorphus:
Perhaps @bucaran has the original SVG?

scorphus:
> The original version is disrespecting the rounding harmony:

[...]

PR #139, opened 2015-10-21
Revamp current logo, use SVG version on README

derekstavis:
This PR replaces the PNG logo by a SVG one, closing #131


bpinto:
We could keep the `.png` file. I thought we might want to add the `.psd` file, but better not.

derekstavis:
The SVG file is the best source for generating raster files. I see no reason to keep the PNG version if we don't use it anywhere :)

bpinto:
We could use. The file is already on git, removing it won't decrease the repository size.

derekstavis:
Ok. Should I update it based on a SVG version rasterisation?

derekstavis:
Here we go!

scorphus:
I have finished polishing the SVG file. These are the differences after all the angles were correctly balanced and curves perfectly adjusted:

diff

I still need to optimize it, remove some unnecessary stuff from it. Please bear with me while I do so. Thanks.

derekstavis:
Be sure that the logo is symmetric! We don't want to swap a crooked logo by another one!

scorphus:
Oh, sure. That's exactly one of the things that it **wasn't**, symmetric. I'll push to this very branch as soon as I'm done. Ok?

derekstavis:
I'd prefer heading to #139 or posting the result here to discuss before pushing to the branch.

scorphus:
Works for me.

derekstavis:
Nice. Thanks! :+1:

scorphus:
@all: Sorry for taking so long. Please, have a look. Thanks!

derekstavis:
Indeed more aligned, looks almost the same of my original Illustrator file (before grid alignment). It happens that you just added the floating point, which I have _intentionally_ removed for alignment and rendering reasons.

Anyway, in the meantime I have wondered about what adjustments could be done in PR's current logo and came with a better version whose alignment is based on iOS icon grid. I have worked harder to ensure [grid aligning](https://helpx.adobe.com/illustrator/using/drawing-pixel-aligned-paths-web.html) by design, instead of a [post-process task](https://github.com/svg/svgo), as I have done the first time.

As soon as I can get hands on my computer I will post it here.

derekstavis:
Here we go. Some design points that were worked out on this version:

- Image size: 1024 x 1024
  - 8's multiple for better resizing
- Pixel alignment by design
  - [Better rendering](http://www.adobepress.com/articles/article.asp?p=1929168&seqNum=5) when resized
- Fish size and position based on a grid
  - [iOS icon grid](http://40.media.tumblr.com/5c09fe1d9c083e93850dfca1d6d6342b/tumblr_moc979zqpS1qcbqvco1_1280.png)
- Background corner rounding: 48px

URLs now use rawgit service to serve the SVG logo with right Content-Type. For preview reasons they currently point to this repo user/branch, but will be adjusted before the merge.

bpinto:
ping @scorphus 

I like that the link is using rawgit which means we can improve it and still change the logo on all READMEs.

bpinto:
:+1:

scorphus:
Remarkable job, @derekstavis! I've just one thing to note. I think these four points **can** be removed without losing **anything**:
clip

Aside from that, it rocks!

BTW, I'm gonna ask my fellow designers to have a look and see what they say about it, maybe they have something to contribute.

Thanks!

derekstavis:
The bigger circled points could be removed. I will give a look at home. The other ones are needed for pixel grid aligment, in a way we don't get broken widths.

PR #143, opened 2015-10-25
add pkg export

jcouyang:


derekstavis:
By OMF convention, the repository name must be prefixed with just "plugin". In this case the correct URL should be `https://github.com/jcouyang/plugin-export`. You also don't need to prepend `.git` :)

derekstavis:
Ah, don't forget to squash your commits into a single one  :shipit:

derekstavis:
Awesome! :ship: it @bpinto!

bpinto:
:+1:

PR #149, opened 2015-11-02
Implement automatic dependency resolution

derekstavis:
This PR adds automatic dependency loading to packages. Fixes #64 

Packages must include a `bundle` file in which dependencies are
declared in the same syntax as `$OMF_CONFIG/bundle` file.

On package installation we install the declared bundle.
On shell startup we load the dependencies from the bundle file.


bpinto:
This is great!
ping @bobthecow

bpinto:
:+1:

PR #150, opened 2015-11-02
autoload: Use `set -e` to remove components from paths

derekstavis:
When using `autoload -e` we currently do a copy of destination function path's content,
modify it and then set it back. This was done as a workaround for using `set -e` with a
precedency issue in `set -e $dest[1]` usage, which is solved by using `set -e {$dest}[1]`.


bpinto:
This is sweeeeeeeeeeeeeeeeet!

PR #171, opened 2015-11-10
git helpers repo

dimitrismistriotis:
Wrote some helpers for git including
 https://twitter.com/paul_irish/status/648248710590148608 and ported aliases from oh-my-zsh git plugin.

### DIFF
--- a/db/pkg/git-helpers
+++ b/db/pkg/git-helpers
@@ -1,0 +1,1 @@
+https://github.com/dimitrismistriotis/git-helpers


bpinto:
Hey @dimitrismistriotis thanks for the PR, we don't have any git helpers on the repository because they are in the end, personal aliases.

So even though we won't merge it, I am sure others will read this PR and install it themselves if they want to. Thanks for making the aliases public. :+1:

bobthecow:
Should we add a wiki page for listing plugins like this?

PR #172, opened 2015-11-12
Suppress basename usage output

bpinto:
When using basename without any arguments, a usage helper as following
is displayed:

```
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
```

When listing the installed packages, if no theme was installed (e.g.
first installation), the `themes/` folder are empty and the usage helper
is displayed. This change addresses it by suppressing the message.


derekstavis:
Great! :+1:

Issue #152, opened 2015-11-03
Support an optional `omf-` prefix for repos?


bobthecow:
When a package repo is in the omf org, it's namespaced: `oh-my-fish/theme-foo`. But when it's a user's repo, it's not: `fred/plugin-bar`. In this case, it's not super obvious what it's a plugin for. We should support an optional `omf-` prefix, which is removed just like the `plugin-` and `theme-` prefixes :)

derekstavis:
I like the idea!

bpinto:
I like the idea too (being optional).

thecotne:
repository name should not matter at all for user repos 
plugin name can be taken from file name in `db/pkg`
or something like package.json can be created in repository or something like this can be done https://github.com/wbond/package_control_channel/blob/master/repository/a.json

i could contribute to this but i don't know fish language and also i can't understand how someone can write actual script in shell scripting languages i can't even use awk in my plugin

bpinto:
> i can't understand how someone can write actual script in shell scripting languages i can't even use awk in my plugin

:joy: 

`package.json` is more complicated than what we need. All we need is to answer a simple question: is it a theme or a plugin?

Simple is better here. Supporting`omf` is a good solution.

thecotne:
if i could write this in programming language i would lookup in db directory and there if repository is in pkg directory then it's plugin if in themes it's theme very simple

plugin/theme name is filename in db directory and repository name is never used anywhere 

i always like simple in use not in implementation (if you can't achieve both)

btw i managed to rewrite php script with awk and also simplify regex https://github.com/thecotne/omf-artisan/commit/c7142c3815099393c1bc40e063931eca8205cae6

i want to try implementing this but how can i test if change in script it worked or not? (debugging/testing)

bpinto:
> if i could write this in programming language i would lookup in db directory and there if repository is in pkg directory then it's plugin if in themes it's theme very simple

You don't have this for plugins that are not on the database and we do support installing from any URL and soon folder on your computer.

> i always like simple in use not in implementation (if you can't achieve both)

It's very simple to use, we just add it to the name of the plugin.

Anyway, I don't think there is any point on discussing this, it's certainly not what I want to do nor you. We both want to improve the framework. So let's focus on how to improve it instead.

---

> i want to try implementing this but how can i test if change in script worked or not? (debugging/testing)

[...]

PR #178, opened 2015-11-23
add artisan

thecotne:


derekstavis:
Hi @thecotne. Thanks for the PR! By default Oh My Fish repository names must start with `plugin-` and themes with `theme-`. Would you mind fixing this issue before we merge this package into db?

thecotne:
@derekstavis why does repository name matter? 
i don't want to change my repository name to `plugin-artisan` because it does not tell anything about repository

it is oh my fish (omf) plugin for artisan so i named it accordingly

i can rename it to omf-plugin-artisan if you want to mention plugin in repository name
but i don't want to remove omf prefix since it is not plugin for everything that exists and may exist in future it is only for omf

also i looked at db and find out that many plugins does not follow this convention there are pkg prefixed and omf prefixed repository's

derekstavis:
> why does repository name matter?

1. Standardization
2. Differentiate different package types
3. It's [a Oh My Fish rule](https://github.com/oh-my-fish/oh-my-fish/blob/master/pkg/omf/cli/omf.install.fish#L22)

> also i looked at db and find out that many plugins does not follow this convention

There are no plugins which does not follow this convention: `basename -a (cat $OMF_PATH/db/pkg/*)`

>... there are pkg prefixed and omf prefixed repository's 

`pkg-` is supported for backwards compatibility.

> i can rename it to omf-plugin-artisan if you want to mention plugin in repository name

There are an open issue about this at #152. I would be glad to accept PRs implementing it.

bpinto:
I totally missed this, is this ready @thecotne ?

thecotne:
@bpinto nope it has `omf-` prefix

bpinto:
:+1:

oranja:
@thecotne **omf-** prefix is supported now. Want to rebase your PR so this can be merged?

bpinto:
Thanks! Plugin added!

PR #181, opened 2015-11-28
readme: Fix wrong link to FAQ

mrshu:
* The English README.md had a link to FAQ that does not work anymore given that it has been localized to `zh-CN` too. This commit makes sure that the link to FAQ in the English article points to the correct part of the correct file.


derekstavis:
Nice! Thanks for the fix!

PR #187, opened 2015-12-03
Add md

starandtina:
Add md plugin to make directories with option -p.


derekstavis:
Hello @starandtina. Thanks for the PR. We don't accept alias packages as they are very personal. You can keep the package into your profile and continue installing it via URL, btw.

PR #188, opened 2015-12-03
Add l

starandtina:
Add l to list contents of directoy with -alh.


derekstavis:
Hello @starandtina. Thanks for the PR. We don't accept alias packages as they are very personal. You can keep the package into your profile and continue installing it via URL, btw.

PR #190, opened 2015-12-04
Better organized 'omf theme' output

oranja:
`omf theme` now lists the current, installed and other available themes in different colors to mark their status.
Current theme is emphasized with `omf::em`,
installed themes are displayed normally,
and other available themes are dimmed with `omf::dim`.
Implementation is a bit hacky, but I didn't want to clutter `omf.fish` with very mini-functions specific only to the `theme` sub-command.
Has been tested on Fish 2.2.0, running on Fedora 23 (x64).
Mac OSX behavior should still be tested before merging to master (different `sed` syntax?)

**Edit:**

1. I fixed the editing errors of the last commit, please reconsider. Sorry for wasting your time with a faulty submission.
2. The regular expression that was supposed to highlight the current theme allowed partial matches, so that when applying the agnoster theme, for example, the token _agnoster_ would be highlighted both in **agnoster** and **agnoster**-mercurial.
3. As previously noted, implementation is somewhat patchy. It assumes that no theme name begins with a '+' or '-' sign. A different approach to distinction of the installed from available themes would be simply to display them separately.

**Edit 2:**
Preview image!
omf_theme_colorful

### DIFF
--- a/pkg/omf/omf.fish
+++ b/pkg/omf/omf.fish
@@ -21,6 +21,10 @@ function omf::dim
   set_color $fish_color_autosuggestion ^/dev/null; or set_color 555
 end
 
+function omf::under
+  set_color --underline
+end
+
 function omf::err
   set_color $fish_color_error ^/dev/null; or set_color red --bold
 end
@@ -119,13 +123,14 @@ function omf -d "Oh My Fish"
 
     case "t" "theme"
       if test (count $argv) -eq 1
-        set -l ostype (uname)
         set -l theme (cat $OMF_CONFIG/theme)
-        set -l regex "[[:<:]]($theme)[[:>:]]"
-        test "$ostype" != "Darwin"; and set regex "\b($theme)\b"
+        set -l highlight_current "(^|[[:space:]])($theme)([[:space:]]|\$)"
 
-        omf.packages.list --database --theme | column | sed -E "s/$regex/"(omf::em)"\1"(omf::off)"/"
-        omf::off
+        echo (omf::under)"Installed:"(omf::off)
+        omf.packages.list --installed --theme | column | sed -E s/"$highlight_current"/"\1"(omf::em)"\2"(omf::off)"\3"/g
+        echo
+        echo (omf::under)"Available:"(omf::off)
+        omf.packages.list --available --theme | column
       else if test (count $argv) -eq 2
         omf.theme $argv[2]
       else


pkg/omf/omf.fish
```diff
@@ -115,16 +119,33 @@ function omf -d "Oh My Fish"
           echo (omf::err)"Argument missing"(omf::off) 1^&2
           echo "Usage: $_ "(omf::em)"$argv[1]"(omf::off)" "(omf::em)"pkg|themes"(omf::off)"/<name> <url>" 1^&2
           return $OMF_MISSING_ARG
+          function omf::under
+            set_color --underline
+      e
```
bpinto: This is broken. :dancers:

oranja: sorry! my touchpad is acting crazy lately :(
will fix soon and reapply

derekstavis: Looks like `omf::under` is being defined twice. Why is it?

bpinto:
Hey that looks good!! I'll look at the code soon!

pkg/omf/omf.fish
```diff
@@ -118,13 +122,26 @@ function omf -d "Oh My Fish"
       end
 
     case "t" "theme"
+      # Note: wasn't tested on Darwin/MacOSX
       if test (count $argv) -eq 1
```
bpinto: Could you remove this? We don't want it on the code base :)

bpinto:
Looks good to me! @derekstavis will double check it later today/tomorrow (time zone)

pkg/omf/omf.fish
oranja: Of course you don't want it in the code base...
I'm trying to see if any of my Mac friends use Fish (shame on them twice if they don't) and can help me test if the changes are compatible with OSX tools. Since I don't trust that any of them will be helpful, I'm already creating an OSX VirtualBox for this check...
I'd prefer to test it properly before removing this line (originally I was hoping that one of you guys uses Mac).

bpinto: We use! 😀

oranja: Ohh, so cool :) it's out.

derekstavis:
I had some issues here with misalignment of output:

screenshot from 2015-12-08 14-46-48

oranja:
Sorry for the long absence. I've been trying to find a workaround for this problem today, but the fact is that the code only gets uglier and uglier and still it doesn't work well. It is possible to reimplement `column` with better handling of escape sequences, but I think that would be an overkill and not welcome in omf's code base.
So instead I'm going to propose a simpler and more straightforward approach. We drop the color coding and print the two categories separately. Hopefully I'll have something to show later today.

oranja:
So it's not really colorful anymore. Even the simplest color coding might break the columns.
Still... I think it's a lot better than getting a list of all themes and getting errors for trying those that are not installed locally.
And for any use case that requires a clean list there's still `omf.packages.list --theme`.
Hope you like it.
omf_theme_notreallycolorfull

bpinto:
Well, it's not as pretty as you said. But it works™.

Looks good to me.

derekstavis:
Maybe making the number of columns fixed would make it better. But yeah, I kinda like it.

oranja:
What you suggest is possible, but I think what you really want is to fix the number **and** the width of columns. That too is not possible with `column`, unfortunately.

This is how it looks like when fixing the number of columns to 3 (with `paste`, how silly):
omf_theme_notreallycolorfull3

Again, another option would be to reimplement `column` locally and bundle that with oh-my-fish.
Even if someone proposed a successful patch for `column` that ignores the non-printables, I don't know how long it takes before the new version is widespread enough. Plus, I'm not sure if OS X is considered downstream for the relevant package (linux-utils).
Before any of that, I guess you should first decide if it's necessary. Pick a TUI style, document it (to make it formal and known to collaborators) and enforce it. If columns and colors are going to be everywhere, it might be worth to implement a home-made `column` function.

derekstavis:
I see, and you are right about my desire. About escape sequences, the solution would be adding escape sequences to every entry (even if the escape sequence is a "reset color" one), this way `column` would work accordingly, iirc.

derekstavis:
Also, instead of using the "Current" approach, highlighting the current on installed would look greater :)

oranja:
I know. I ditched it because I saw it's coming back to the same misalignment problem with colors.

The workaround you suggest might actually work. I never thought of it. Now the question is how would you like to see it eventually? Explicit categories or color codes (preferably with a legend)?

derekstavis:
I prefer categories. It's more explicit, more clear and more organized.

oranja:
Last preview (I hope):
omf_theme_notreallycolorfull4

derekstavis:
I love the result! Would you mind rebasing the commits, maybe even squashing into a single one?

pkg/omf/omf.fish
```diff
@@ -21,6 +21,10 @@ function omf::dim
   set_color $fish_color_autosuggestion ^/dev/null; or set_color 555
 end
 
+function omf::under
```
derekstavis: This isn't being used anymore. I liked underline for section titles, better than `omf::dim`

pkg/omf/omf.fish
```diff
@@ -119,13 +123,19 @@ function omf -d "Oh My Fish"
...
-        set -l regex "[[:<:]]($theme)[[:>:]]"
-        test "$ostype" != "Darwin"; and set regex "\b($theme)\b"
-
-        omf.packages.list --database --theme | column | sed -E "s/$regex/"(omf::em)"\1"(omf::off)"/"
-        omf::off
+        # regex for highlighting the current theme
+        set -l theme_regex "(^|\s)($theme)(\s|\$)"
+        # regex for wrapping everything else with (omf::off), just in order to keep the alignment
+        set -l others_regex "\S+"
+
```
derekstavis: I'd use the more semantically names: `current_theme_regex` and `theme_regex`.

pkg/omf/omf.fish
```diff
@@ -119,13 +123,19 @@ function omf -d "Oh My Fish"
...
       if test (count $argv) -eq 1
-        set -l ostype (uname)
         set -l theme (cat $OMF_CONFIG/theme)
-        set -l regex "[[:<:]]($theme)[[:>:]]"
-        test "$ostype" != "Darwin"; and set regex "\b($theme)\b"
-
-        omf.packages.list --database --theme | column | sed -E "s/$regex/"(omf::em)"\1"(omf::off)"/"
-        omf::off
+        # regex for highlighting the current theme
+        set -l theme_regex "(^|\s)($theme)(\s|\$)"
```
derekstavis: change to `set -l highlight_current`, then no comment is needed.

pkg/omf/omf.fish
```diff
@@ -119,13 +123,19 @@ function omf -d "Oh My Fish"
...
+        # regex for wrapping everything else with (omf::off), just in order to keep the alignment
+        set -l others_regex "\S+"
```
derekstavis: `set -l normal_item` better?

oranja:
Sorry for all this going back and forth. I think it's done now.
@derekstavis thanks for the helpful review.
@bpinto thank you for your time and feedback.

I'd appreciate it if you guys verified that it still works on Mac OS X. I don't have the means to do it myself.

As for the style, it now looks exactly as in the previous screenshot, except that the Installed and Available headers are now underlined and not dimmed.

bpinto:
Hey @oranja thank you! It was a great addition! And we wouldn't have done without you!

oranja:
Seems like I still have much to learn on git rebasing. I hope I'm not screwing anything up!

bpinto:
It looks like the rebase was not complete.

This is how I'd do:

```
git checkout master
git pull --rebase upstream/master
git checkout omf-theme-colorful
git rebase master
```

oranja:
Thanks. Simple and works great.

bpinto:
There is something funny happening with the color:

With one theme installed:
screen shot 2015-12-18 at 15 44 03

With more than one theme installed:
screen shot 2015-12-18 at 15 44 25
screen shot 2015-12-18 at 15 44 51

oranja:
It works just fine here with one and two installed themes, so I suspect it's a Mac thing.
Does it happen with more themes installed as well?
Is it GNU sed? I'm looking at Mac's `man` page for [`sed`] and it looks like the regexes should work well enough.
Do you have any other ideas why this is happening?

[`sed`]: https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/sed.1.html

bpinto:
I'll look into it again and find the fix when I get home!

pkg/omf/omf.fish
```diff
@@ -119,13 +123,18 @@ function omf -d "Oh My Fish"
...
     case "t" "theme"
       if test (count $argv) -eq 1
-        set -l ostype (uname)
         set -l theme (cat $OMF_CONFIG/theme)
-        set -l regex "[[:<:]]($theme)[[:>:]]"
-        test "$ostype" != "Darwin"; and set regex "\b($theme)\b"
-
-        omf.packages.list --database --theme | column | sed -E "s/$regex/"(omf::em)"\1"(omf::off)"/"
-        omf::off
+        set -l highlight_current "(^|\s)($theme)(\s|\$)"
```
bpinto: Doesn't this mean start and end of a line?

pkg/omf/omf.fish
```diff
@@ -119,13 +123,18 @@ function omf -d "Oh My Fish"
...
+        # regex for wrapping everything else with (omf::off), just in order to keep the alignment
+        set -l normal_item "\S+"
+
+        echo (omf::under)"Installed:"(omf::off)
+        # align before adding color codes. otherwise column messes the alignment
+        # this marks the current theme twice, both with 'off' and 'dm', but this "mistake" actually keeps the alignment better for some reason
+        omf.packages.list --installed --theme | column | sed -E s/"$highlight_current"/"\1"(omf::em)"\2"(omf::off)"\3"/g | sed -E s/"$normal_item"/(omf::off)"\0"(omf::off)/g
```
bpinto: I find it weird that you need to parse both the default theme and the others. Why is it necessary now?

pkg/omf/omf.fish
oranja: It should be a word boundary, only that `\b` also matches hyphens (-), so that choosing _agnoster_ highlighted both **agnoster** and **agnoster**-mercurial.
So it's  -- either the beginning of the line - or - a whitespace character -+- the name of the theme -+- a whitespace character - or - the end of the line --.
And of course `$` marks the end of the line but has to be escaped in the shell so that Fish doesn't expect a variable name.

pkg/omf/omf.fish
oranja: See @derekstavis's screenshot above (one of the first in this PR). Alignment with highlighting was a tough nut and so far that's the only way I managed to make it work. Spent way too many hours and eventually the solution is a ugly hack. Maybe one day I will write a `column` alternative in Fish...

pkg/omf/omf.fish
bpinto: In order to fix this on OSX, I use this regex instead: `set -l highlight_current "(^|[^\w\-])($theme)(\$|[^\w\-])"`

pkg/omf/omf.fish
bpinto: This does not work?
`omf.packages.list --installed --theme | column | sed -E s/"$highlight_current"/"\1"(omf::em)"\2"(omf::off)"\3"/g`

oranja: Guess what... now it seems to work. I was probably messing around with too many variables to notice that adding the white-spaces solves the problem. :grin:

pkg/omf/omf.fish
oranja: Does changing the order of the `$` and `\s` help anything? I know it probably shouldn't... Just hoping.
My problem with this regex is that there is no explicit limit on package names (except for the prefixes for repositories). I guess no dev would try to add spaces to a package name, but what if I wanted to be hip and call my theme "w~a~v~e" (or anything like that). This regex will stop highlighting after the 'w'.
Maybe it's for the better to limit package naming possibilities, but I wouldn't allow it in a script before it's documented.

bpinto: The order doesn't change anything. What I have seen is that there is no '\s' for osx. We would have to use a similar regex we used to that was exclusive for osx.

I think documenting is a good start! Could you update the documentation?

pkg/omf/omf.fish
bpinto: 😆👍

pkg/omf/omf.fish
oranja: But.. but.. `\s` is right there in [re_format]'s man page. :disappointed: 
Can you try now? `[[:space:]]` seems to work for GNU sed and it should work for OS X too.

[re_format]: https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man7/re_format.7.html

bpinto:
I also read the manual and it does mention as if it had support. I have either missed a key word or the manual page is just wrong.

Yesterday while working on the regex, I found some stackoverflow comments saying the manual was wrong on some osx versions. 😕

bpinto:
I'll check this again tomorrow, sorry I didn't do it today.

bpinto:
:+1: 
This new regex works great here!

pkg/omf/omf.fish
```diff
@@ -119,13 +123,18 @@ function omf -d "Oh My Fish"
...
-        omf.packages.list --database --theme | column | sed -E "s/$regex/"(omf::em)"\1"(omf::off)"/"
-        omf::off
+        set -l highlight_current "(^|[[:space:]])($theme)([[:space:]]|\$)"
+        # regex for wrapping everything else with (omf::off), just in order to keep the alignment
+        set -l normal_item "\S+"
+
+        echo (omf::under)"Installed:"(omf::off)
+        # align before adding color codes. otherwise column messes the alignment
+        # this marks the current theme twice, both with 'off' and 'dm', but this "mistake" actually keeps the alignment better for some reason
+        omf.packages.list --installed --theme | column | sed -E s/"$highlight_current"/"\1"(omf::em)"\2"(omf::off)"\3"/g # | sed -E s/"$normal_item"/(omf::off)"\0"(omf::off)/g
```
bpinto: I'd remove the code that is no longer necessary (remove the comment), and redo the line comment above a it's outdated.

pkg/omf/omf.fish
```diff
@@ -119,13 +123,18 @@ function omf -d "Oh My Fish"
...
-        set -l ostype (uname)
         set -l theme (cat $OMF_CONFIG/theme)
-        set -l regex "[[:<:]]($theme)[[:>:]]"
-        test "$ostype" != "Darwin"; and set regex "\b($theme)\b"
-
-        omf.packages.list --database --theme | column | sed -E "s/$regex/"(omf::em)"\1"(omf::off)"/"
-        omf::off
+        set -l highlight_current "(^|[[:space:]])($theme)([[:space:]]|\$)"
+        # regex for wrapping everything else with (omf::off), just in order to keep the alignment
+        set -l normal_item "\S+"
```
bpinto: No longer used.

pkg/omf/omf.fish
```diff
@@ -119,13 +123,18 @@ function omf -d "Oh My Fish"
...
     case "t" "theme"
       if test (count $argv) -eq 1
-        set -l ostype (uname)
         set -l theme (cat $OMF_CONFIG/theme)
-        set -l regex "[[:<:]]($theme)[[:>:]]"
-        test "$ostype" != "Darwin"; and set regex "\b($theme)\b"
-
-        omf.packages.list --database --theme | column | sed -E "s/$regex/"(omf::em)"\1"(omf::off)"/"
-        omf::off
+        set -l highlight_current "(^|[[:space:]])($theme)([[:space:]]|\$)"
```
bpinto: Being able to use the same regex is great!

oranja:
@bpinto Thanks for helping me test on OSX.
Sorry about the leftovers from before. It's clean now.

derekstavis:
Awesome. Merged! :+1:

Issue #185, opened 2015-11-30
Manual installation instructions


jcelliott:
It would be nice to have some simple manual installation instructions. I already have a lot of configuration around fish, and I don't like that the install script tries to replace everything. The old Oh My Fish style setup seemed to be a lot more friendly to existing configurations. Any suggestions on improving this?

rrosson:
What I did was backup my ~/.config directory and then removed my original fish directory and related files from .config. Installed OMF then added my customizations to ~/.config/omf/init.fish

Hope this helps.

derekstavis:
Thanks for sharing your experience with Oh My Fish. I will try to comment your points:

>It would be nice to have some simple manual installation instructions.

Currently you can do a manual install using smth like this:

```fish
cd <WHERE YOU WANT OMF TO BE>
git clone https://github.com/oh-my-fish/oh-my-fish
echo 'set -gx OMF_PATH '(pwd) >> ~/.config/fish/config.fish
echo 'source $OMF_PATH' >> ~/.config/fish/config.fish
```
> I already have a lot of configuration around fish, and I don't like that the install script tries to replace everything

We don't replace, we do a backup of your current configuration.

> The old Oh My Fish style setup seemed to be a lot more friendly to existing configurations. Any suggestions on improving this?

I would to hear your suggestions on this :dancers:

coderstephen:
Perhaps we could be a bit smarter about migration? Maybe run an interactive install wizard that asks some questions like these:

- Would you like to backup `~/.config/fish/config.fish`?
- Would you like to move `~/.config/fish/config.fish` to `~/.config/omf/init.fish`?
  - If not, just append the required bootstrap commands to `~/.config/fish/config.fish`.
- Would you like to move `~/.config/fish/functions` to `~/.config/omf/functions`?

Maybe there's better questions (more or less) we could ask.

bpinto:
It would indeed be nice to have an interactive wizard (with the option to bypass it), go for it @coderstephen :+1:

rrosson:
The only problem I see with the questions is the user has to understand the customizations they have before OMF and to know what packages from OMF my cause conflicts and issues. This would mean there would have to be logic in OMF to check to see if any package would have an issue with the users customizations.

Just thinking out loud.

coderstephen:
Most of these improvement ideas for the installer have been implemented in #361.

PR #192, opened 2015-12-06
Faster CI builds

bpinto:
Split the Dockerfile in two: Fish shell image and Oh My Fish image.
In order to simplify this `.travis.yml` file, we need to add support for [Manual installation](https://github.com/oh-my-fish/oh-my-fish/issues/185).

This has not only decreased the build time considerably, but also allow us to have a stabler build as our fish image is always cached.


derekstavis:
LGTM :+1: (for Travis too)

PR #203, opened 2015-12-19
Better test output

bpinto:
Success:
```
⋊> omf/pkg/fish-spec on better-test-output ⨯ fish-spec
.........

Finished in 0.105 seconds
```

Error:
```
⋊> omf/pkg/fish-spec on better-test-output ⨯ fish-spec
........

Failure: It succeeds when single assertion succeeds
 Expected error to equals success


Finished in 0.101 seconds
```


derekstavis:
LGTM :+1:

PR #205, opened 2015-12-21
Run tests on osx

bpinto:
Run tests on OSX and Linux (ubuntu).

This PR has code from https://github.com/oh-my-fish/oh-my-fish/pull/203 – we are going to merge that one first.


derekstavis:
Looks good, but could be made better by compiling fish from source tarballs, which would allow the build to use the container infrastructure (faster builds on Linux), removing the need of `sudo`, homogeneous "install" process (no testing if Linux or OS X and waiting for Homebrew/APT) and the most important thing: Caching the precompiled/installed stuff.

bpinto:
> compiling fish from source tarballs, which would allow the build to use the container infrastructure (faster builds on Linux)

I'm using the fastest and newest build machines which also uses the latest long-term support Ubuntu (14.04).

Reference: https://docs.travis-ci.com/user/ci-environment/

P.S.: The build time is not accurate as it's always faster than what it is described on the table (link above). Plus, we are running the tests on OS X, so even if the boot time was true, the OSX tests still would take longer to run.

> homogeneous "install" process (no testing if Linux or OS X and waiting for Homebrew/APT)

Fish has several dependencies which need to be installed:

```
+addons:
 +  apt:
 +    packages:
 +    - bc
 +    - doxygen
 +    - expect
 +    - gettext
 +    - libncurses5-dev
 +    - tree
```

We need to know which system we are running (Linux/OSX) and install them.

> and the most important thing: Caching the precompiled/installed stuff.

First, caching is a bonus. We are talking about a build that is taking 50 seconds to complete (check travis). This is not a show stopper as it's a bonus.

Second, the goal of this PR is to run the tests on OSX, not to make it blazing fast. I'd agree if the build was slow, which is not the case.Divide and conquer.

derekstavis:
Looks good to me! :+1:

PR #207, opened 2015-12-23
Reorganize `omf` plugin structure

derekstavis:
This updates omf plugin to new architecture, moving
functions to `functions` directory. It also revamps the
separation of concerns between CLI and underlying
implementations, organizing them into directories,
whose are autoloaded in plugin startup.


pkg/omf/functions/omf.fish
```diff
@@ -0,0 +1,68 @@
...
+    set command $argv[1]
+    set arguments
+  else
+    set command help
+    set arguments
+  end
+
+  switch "$command"
+    case "-v*" "--v*"
+      omf.version
```
bpinto: Omf.cli.version

pkg/omf/functions/omf.fish
```diff
@@ -0,0 +1,68 @@
...
+
+    case "doctor"
+      omf.doctor
```
bpinto: Omf.cli.doctor

rrosson:
The only issue I see with this is what if user defined completions and functions are already in ~/.config/omf? I think in the past this was the recommended place to put these directories and have them called out from the init.fish in the same directory.

If the plan is to move forward this will need to be considered.

pkg/omf/functions/packages/omf.packages.remove.fish
```diff
@@ -0,0 +1,46 @@
...
+      or continue
+
+    test $pkg = (cat $OMF_CONFIG/theme);
+      and echo default > $OMF_CONFIG/theme
+
+    if rm -rf $path
+      omf.bundle.remove "theme" $pkg
+      return 0
+    else
+      return 1
```
bpinto: We don't use these return value.

pkg/omf/functions/packages/omf.packages.update.fish
```diff
@@ -1,11 +1,14 @@
-function omf.update -a name
-  function __omf.update.success
-    echo (omf::em)"✔ $argv successfully updated."(omf::off)
-  end
+function __omf.packages.update.success
+  echo (omf::em)"✔ $argv successfully updated."(omf::off)
+  return 0
```
bpinto: Everywhere else we have these helper functions we dont put the return value inside of the function.

pkg/omf/functions/packages/omf.packages.remove.fish
derekstavis: They are now used on cli to distinguish package not found from rm failing

pkg/omf/functions/packages/omf.packages.update.fish
derekstavis: As they are the last function called by parent function their return values are propagated

pkg/omf/omf.bundle.install.fish
```diff
@@ -0,0 +1 @@
+functions/bundle/omf.bundle.install.fish
```
bpinto: ?

derekstavis: It's a symlink :)

derekstavis:
Upstreaming! :shipit:

Issue #196, opened 2015-12-11
System package dependencies


coderstephen:
Now that we have package dependencies, we should add a feature to check for system dependencies as well for packages. For example, the weather plugin depends on the shark plugin, but it also depends on `jq` being installed. I propose enhancing the bundle file to iterate over `available` requirements as well, which would be verified by using the `available` function.

Package bundle files could now look like this:

    available jq
    package shark

If one or more `available` dependencies are not satisfied, the package should not be installed and the missing dependencies should be listed. We shouldn't try to install them ourselves.

We'll have to go through the whole file first to test the `available` lines first, since we shouldn't install any dependent OMF packages either if the dependencies aren't satisfied.

Thoughts?

derekstavis:
:+1: I like the idea! It's awesome! Better than implementing hand check over `init.fish`

coderstephen:
Here's a proof-of-concept that I have working: https://github.com/coderstephen/oh-my-fish/tree/feature/available-dependencies

oranja:
Haven't tested it but it looks good at first glance.

Say, does `available` really work with a URL?
I have _shark_ installed. `available shark` returns 0, but `available (cat db/pkg/shark)` returns 1.
Does it matter if I installed it from OMF's db or an external URL?
Anyway, I don't think it should be possible to specify a package by a full URL. Just a name and optionally a minimal version number. Not that versions are fun... It sucks when two packages want two different versions of the same dependency. :confounded:

coderstephen:
@oranja No, and it isn't meant to. We already have a `package` directive to depend on actual packages (which also works with external URLs). `available` is meant for system dependencies that we actually are unable to install ourselves. For example, if a package uses OpenSSL commands, it should have

    available openssl

in its bundle file. OMF can't install OpenSSL, but we can at least warn the user that OpenSSL should be installed first before the package will work.

bpinto:
Hey @coderstephen

I think It might be better to add support for this inside a package.json-similar way. What do you think?

We could start a discussion around this configuration file.

derekstavis:
>I think It might be better to add support for this inside a package.json-similar way. What do you think?

Yes. I agree with `package.json`. What comes to my mind is something like this:

```fish
$ cat package
name: balias
type: theme
description: a description override (default is from github repo)
```

bpinto:
I'm not sure if we should keep both bundle and package file. Shouldn't we merge them?

oranja:
[...]

Issue #191, opened 2015-12-05
Make theme- prefix no longer required for theme packages


edouard-lopez:
_Original title: How to install 3rd-party theme?_

---

I tried to install the fish-port of [`pure` theme](https://github.com/rafaelrinaldi/pure) without success. I tried the following commands:

```
omf install https://github.com/rafaelrinaldi/pure; omf theme pure
omf install https://github.com/rafaelrinaldi/pure.git; omf theme pure
omf install https://raw.github.com/rafaelrinaldi/pure/master/fish_prompt.fish; omf theme pure
```

all failed. What is the correct way to do that?

bpinto:
Hi, I'm sorry but the package name does not conform to the rules. In order for omf to understand that the package is a theme, it needs to have a prefix theme-

So you would have to rename the repository from pure to theme-pure.

bobthecow:
it seems like, for third party url-based installs, we should support no prefix at all as well.h

derekstavis:
I agree, then we would need to support:

```fish
omf install --theme URL
omf install --plugin URL
```

bobthecow:
Yep. Have it work as currently implemented for `omf install`, and if there's no prefix have it prompt the user to install with `--theme` or `--plugin`

edouard-lopez:
Making the `theme-` prefix would solve [our situation](https://github.com/rafaelrinaldi/pure/issues/3) where we are wondering if you need to submodule the theme in a fork repo just to be omf compatible.

bucaran:
Hi! @edouard-lopez Check out [Fisherman](https://github.com/fisherman/fisherman), there you can install any themes or plugins (including OMF) without restrictions.

```fish
fisher install rafaelrinaldi/pure
```

bucaran:
https://github.com/oh-my-fish/oh-my-fish/issues/191

Hi! @edouard-lopez Check out [Fisherman](https://github.com/fisherman/fisherman), there you can install any themes or plugins (including OMF) without restrictions.

bucaran:
Hi! @edouard-lopez You can also check out [Fisherman](https://github.com/fisherman/fisherman), there you can install any themes or plugins (including OMF) without restrictions.

```fish
fisher install rafaelrinaldi/pure
```

bucaran:
Hi! @edouard-lopez You can also check out [Fisherman](https://github.com/fisherman/fisherman), there you can install any themes or plugins (including OMF) without restrictions.

```fish
[...]

PR #210, opened 2015-12-25
Support omf- prefix for packages

oranja:
As requested in #152.
Does not solve #191. Prefer to infer package type from metadata in `package` file, as suggested in #196.

### DIFF
--- a/pkg/omf/cli/omf.bundle.remove.fish
+++ b/pkg/omf/cli/omf.bundle.remove.fish
@@ -10,12 +10,11 @@ function omf.bundle.remove
 
       for record in $bundle_contents
         set record_type (echo $record | cut -d' ' -f1)
-        set record_name (echo $record | cut -d' ' -f2-)
-        set record_basename (basename (echo $record_name | \
-          sed -e 's/\.git$//') | sed 's/^pkg-//;s/^plugin-//;s/^theme-//')
+        set record_name_or_url (echo $record | cut -d' ' -f2-)
+        set record_name (omf.package_name $record_name_or_url)
 
-        if not test "$type" = "$record_type" -a "$name" = "$record_basename"
-          echo "$record_type $record_name" >> $bundle
+        if not test "$type" = "$record_type" -a "$name" = "$record_name"
+          echo "$record_type $record_name_or_url" >> $bundle
         end
       end
     end
--- a/pkg/omf/cli/omf.bundle.install.fish
+++ b/pkg/omf/cli/omf.bundle.install.fish
@@ -16,7 +16,7 @@ function omf.bundle.install
       set name_or_url (echo $record | cut -s -d' ' -f2- | sed 's/ //g')
       test -n "$name_or_url"; or continue
 
-      set name (basename $name_or_url | sed 's/\.git//;s/^pkg-//;s/^plugin-//;s/^theme-//')
+      set name (omf.package_name $name_or_url)
 
       if not contains $name $packages
         omf.install $name_or_url;
--- a/pkg/omf/cli/omf.install.fish
+++ b/pkg/omf/cli/omf.install.fish
@@ -17,9 +17,9 @@ function omf.install -a name_or_url
 
   if test -e $OMF_PATH/db/$parent_path/$name_or_url
     set name $name_or_url
-    set url (cat $OMF_PATH/db/$parent_path/$name_or_url)
+    set url (cat $OMF_PATH/db/$parent_path/$name)
   else
-    set name (basename $name_or_url | sed 's/^pkg-//;s/^plugin-//;s/^theme-//;s/\.git$//')
+    set name (omf.package_name $name_or_url)
     set url $name_or_url
   end
 
--- a/pkg/omf/util/omf.package_name.fish
+++ b/pkg/omf/util/omf.package_name.fish
@@ -1,0 +1,3 @@
+function omf.package_name -a name_or_url
+  basename $name_or_url | sed -E 's/^(omf-)?((plugin|pkg|theme)-)?//;s/.git$//'
+end


pkg/omf/util/omf.util_bundle_name.fish
```diff
@@ -0,0 +1,3 @@
+function omf.util_bundle_name -a name_or_url
```
derekstavis:
```fish
function omf.sanitize_package_name -a name_or_url
```

pkg/omf/cli/omf.bundle.remove.fish
```diff
@@ -1,24 +1,8 @@
...
-        set record_type (echo $record | cut -d' ' -f1)
-        set record_name (echo $record | cut -d' ' -f2-)
-        set record_basename (basename (echo $record_name | \
-          sed -e 's/\.git$//') | sed 's/^pkg-//;s/^plugin-//;s/^theme-//')
-
-        if not test "$type" = "$record_type" -a "$name" = "$record_basename"
-          echo "$record_type $record_name" >> $bundle
-        end
-      end
+      sed -i'' /'^type[[:space:]]+name$'/d $bundle
```
derekstavis: There are differences in GNU and BSD `sed -i` usage. Prefer not using `sed -i`.

derekstavis:
How does the changes to `omf.bundle.remove` affect the behavior of how bundle works?

derekstavis:
I prefer not changing the bundle implementation in this PR. Although, nice idea in generalising the name sanitisation function. DRY treatment :+1:

pkg/omf/util/omf.util_bundle_name.fish
oranja: I prefer bundle_name or bundlename, as in basename, just because sanitize has unfortunate associations in my mind. Plus, at this point it is somewhat misleading to claim that the function seriously deals with any security concerns.
Anyway, these are my 2 cents. You're the architect, so say the word and it's changed.

derekstavis: In this line I prefer `omf.package_name`. Bundle is a different story. Bundle is just a file where packages installed by user are kept for dotfile means.

oranja:
omf.bundle.remove:
* You're right about `-i`, but luckily this `sed` issue can be dealt with.
* The changes do not change the behavior of the `bundle`. The previous code removes a specific line from the file. The new code does just that.
* Now I see that I made a dire mistake in the regex, but it's a tiny change to fix it.

Care to elaborate on the rejects? You wish to split the PR or you don't want it altogether?
Except for the `remove` rewrite, which can be skipped, this PR is nothing without the changes to bundle functions.

On the whole `bundle`/`package` confusion, it's tactic. I can go over it again and make sure it fits your definitions.

derekstavis:
About the bundle remove function, if you treat the function as a black box it doesn't do the same thing. Current function actually does three things:

1. Remove the requested record
2. Remove duplicates
3. Sort the bundle file alphabetically

Although arguably more bureaucratic, for traceability reasons I prefer changes unrelated to PR title proposed as separate PRs.

derekstavis:
>On the whole bundle/package confusion, it's tactic. I can go over it again and make sure it fits your definitions.

I don't understand what you mean by tactic. To clarify current terms meanings:

- Packages
  * Plugins
  * Prompt Themes
  * Color Schemes (TBD)
- Bundle
  * A database of installed packages for dotfiles usage
  * A way of declaring package dependencies (soon to be deprecated by metadata)

pkg/omf/cli/omf.bundle.add.fish
```diff
@@ -1,6 +1,6 @@
-function omf.bundle.add -a type name_or_url
```
derekstavis: still `name_or_url`, as when installing from URLs they are used in bundle

pkg/omf/cli/omf.bundle.add.fish
```diff
@@ -1,6 +1,6 @@
...
+function omf.bundle.add -a type name
```
derekstavis: still `name_or_url`, as when installing from URLs they are used in bundle

pkg/omf/cli/omf.bundle.remove.fish
```diff
@@ -11,10 +11,8 @@ function omf.bundle.remove
       for record in $bundle_contents
         set record_type (echo $record | cut -d' ' -f1)
         set record_name (echo $record | cut -d' ' -f2-)
-        set record_basename (basename (echo $record_name | \
-          sed -e 's/\.git$//') | sed 's/^pkg-//;s/^plugin-//;s/^theme-//')
```
derekstavis: I think this line should be
```fish
set record_basename (omf.package_name $record_name)
```

derekstavis:
I think the bulk of this PR should be implementing `omf.package_name` and applying it where regexes was previously used. In the current state it is changing the behaviour of bundle file, mainly when installing packages from URL (bundle is persisting the name, not the URL, as master does).

pkg/omf/cli/omf.bundle.remove.fish
oranja: You're right. Forgot this one when I split the PR.

oranja:
You favor separation, I can't really argue against. I will split the PR when I have the time.
However, I want to be clear that I'm strongly against accepting one part of the PR without the other. It means the same package can be installed from two different sources at the same time. The bundle file should really not care where each package was installed from, and a user should definitely not have to type the full URL to remove a package.

derekstavis:
If we had tests for bundle things would be easier :tea:

Let me draw the two install cases we currently handle and how they are persisted in bundle:

```fish
$ omf install apt
$ cat $OMF_CONFIG/bundle
package apt
$ omf remove apt
$ cat $OMF_CONFIG/bundle
```

```fish
$ omf install https://github.com/derekstavis/plugin-nvm.git
$ cat $OMF_CONFIG/bundle
package https://github.com/derekstavis/plugin-nvm
$ omf remove nvm
$ cat $OMF_CONFIG/bundle
```

The important thing is: When installed via URL, we persist the URL in the bundle, but the package is called by its name (see `omf remove nvm`).

oranja:
>> It means the same package can be installed from two different sources at the same time.

> Not true, `git clone` will fail because the target directory will be the same.

Actually, my bad. It will fail even before `git clone`. Thankfully there's a check in place.


>> The bundle file should really not care where each package was installed from

> Yeah, it should.

At first I was quite displeased that you didn't provide an explanation, but I thought about it some more and you're right. There is at least one use case I could think about, and that is the option to share a `bundle` file between several users / environments and expect exactly the same packages to be installed. Two gotchas there might be different in `db` entries and package versions, but I guess we're not there yet.

derekstavis:
>At first I was quite displeased that you didn't provide an explanation

This is something I have to improve on me. Things are very clear on my mind, but not necessarily for others. Combine this with a bit of laziness and slowness to write and the confusion is done :stuck_out_tongue_closed_eyes:

oranja:
There. Finally had the time to sort out this PR on my side.

derekstavis:
Aww, that looks very good to me!! Thanks for the work :+1: :+1: :+1:

bpinto:
Looking good to me too!

bobthecow:
:+1:

Issue #208, opened 2015-12-24
Shallow git clones for packages


oranja:
The way things are at this point, installing a package with OMF is more or less `git clone`ing a repository from github. This is convenient for implementation, very common and allows easy package updates.
As the title indicates, I'm thinking why not do a small change towards shallow clones (`git clone --depth 1`)?
The benefits are probably negligible in terms of server load, but for the end user it helps to save bandwidth, disk space and time, even if right now most omf packages are only one or two commits.
Any end-user who wants to have the full history will find the trivial way (`git pull --unshallow`). Moreover, it is possible for a developer to `push` changes on top of a shallow clone*
For most, I believe having the latest version is enough. For that matter it is possible\* to `pull` into a shallow clone (and keep it shallow).

Please share your pros and cons if you have any.
Does anyone have a git version earlier than 1.9? :astonished: 

\* since git 1.9

bpinto:
I believe previous osx used to come with 1.8.7, so we would need to support it.

oranja:
OS X... always the party pooper

(oh no, I'm sorry. I take it back! Let's not start a flame war)

derekstavis:
We already do shallow clone on installer. So, not a issue using it for packages.

PR #220, opened 2016-01-02
omf update: Report up-to-date packages as such

oranja:
The purpose of this proposal is to generate a more accurate output on `omf update`. Namely, packages which have no upstream changes now report that they are already up to date.
Since this is a proposal we haven't really discussed before, this is broken into 3 commits (instead of squashed), so it's easier to track the changes and their aims.

Regarding performance, I must admit that I did not time the new code or the old.
Overall the update process should not improve and not deteriorate in terms of performance. The previous `git pull` is broken apart to a `fetch` and a `merge`. Packages without changes skip the merge (and possibly stashing) in return for an additional check. Packages with changes pay just one more check. If `git` does any more optimizations in combining the `fetch` and `merge` when running `pull`, I am unaware of such and I would appreciate any tips.

\* Note that the bugfix is also covered by @derekstavis in **wip/reorganize-code**. I just couldn't leave it like this.

### DIFF
--- a/pkg/omf/functions/packages/omf.packages.update.fish
+++ b/pkg/omf/functions/packages/omf.packages.update.fish
@@ -4,15 +4,20 @@ function omf.packages.update -a name
     not test -e $target_path/.git;
       and return 0
 
-    if omf.repo.pull $target_path
-      omf.bundle.install $target_path/bundle
-      echo (omf::em)"$name successfully updated."(omf::off)
-      return 0
-    else
-      echo (omf::err)"Could not update $name."(omf::off) 1^&2
+    omf.repo.pull $target_path
+    switch $status
+      case 0
+        omf.bundle.install $target_path/bundle
+        echo (omf::em)"$name successfully updated."(omf::off)
+        return 0
+      case 1
+        echo (omf::err)"Could not update $name."(omf::off) 1>&2
+      case 2
+        echo (omf::dim)"$name is already up-to-date."(omf::off)
+        return 0
     end
   else
-    echo (omf::err)"Could not find $name."(omf::off) 1^&2
+    echo (omf::err)"Could not find $name."(omf::off) 1>&2
   end
 
   return 1
--- a/pkg/omf/functions/repo/omf.repo.pull.fish
+++ b/pkg/omf/functions/repo/omf.repo.pull.fish
@@ -1,61 +1,63 @@
 function omf.repo.pull
 
   if test (count $argv) -eq 0
-    echo (omf::err)"Argument of omf.repo.pull is the repo path."(omf::off)
+    echo (omf::err)"omf.repo.pull takes a repository path as an argument."(omf::off) 1>&2
     return $OMF_MISSING_ARG
   end
 
   set -l repo_dir $argv[1]
 
-  function omf.repo.git -V repo_dir
-    command git --git-dir "$repo_dir"/.git --work-tree "$repo_dir" $argv
-  end
-
-  function omf.repo.git.stash -V repo_dir
-    command git -C "$repo_dir" stash $argv
+  function __omf.repo.git -V repo_dir
+    command git -C "$repo_dir" $argv
   end
 
   set -l remote origin
-  if test (omf.repo.git config --get remote.upstream.url)
+  if test (__omf.repo.git config --get remote.upstream.url)
     set remote upstream
   end
 
-  set initial_branch (omf.repo.git symbolic-ref -q --short HEAD);
-    or return $OMF_UNKNOWN_ERR
-  set initial_revision (omf.repo.git rev-parse -q --verify HEAD);
-    or return $OMF_UNKNOWN_ERR
+  set initial_branch (__omf.repo.git symbolic-ref -q --short HEAD);
+    or return 1
+  set initial_revision (__omf.repo.git rev-parse -q --verify HEAD);
+    or return 1
 
-  if not omf.repo.git diff --quiet
+  # the refspec ensures that '$remote/master' gets updated
+  set -l refspec "refs/heads/master:refs/remotes/$remote/master"
+  __omf.repo.git fetch --quiet $remote $refspec;
+    or return 1
+
+  if test (__omf.repo.git rev-list --count master...FETCH_HEAD) -eq 0
+    return 2
+  end
+
+  if not __omf.repo.git diff --quiet
     echo (omf::em)"Stashing your changes:"(omf::off)
-    omf.repo.git status --short --untracked-files
-    omf.repo.git.stash save --include-untracked --quiet;
+    __omf.repo.git status --short --untracked-files
+    __omf.repo.git stash save --include-untracked --quiet;
       and set stashed
   end
 
   if test "$initial_branch" != master
-    omf.repo.git checkout master --quiet
+    __omf.repo.git checkout master --quiet
   end
 
-  set -l refspec "refs/heads/master:refs/remotes/$remote/master"
-
-  # the refspec ensures that '$remote/master' gets updated
-  if not omf.repo.git pull --ff-only --quiet $remote $refspec
-    omf.repo.git checkout $initial_branch
-    omf.repo.git reset --hard $initial_revision
-    set -q stashed; and omf.repo.git.stash pop
+  if not __omf.repo.git merge --ff-only --quiet FETCH_HEAD
+    __omf.repo.git checkout $initial_branch
+    __omf.repo.git reset --hard $initial_revision
+    set -q stashed; and __omf.repo.git stash pop
+    return 1
   end
 
   if test "$initial_branch" != master
-    omf.repo.git checkout $initial_branch --quiet
+    __omf.repo.git checkout $initial_branch --quiet
   end
 
   if set -q stashed
-    omf.repo.git.stash pop --quiet
+    __omf.repo.git stash pop --quiet
 
     echo (omf::em)"Restored your changes:"(omf::off)
-    command git -C "$repo_dir" status --short --untracked-files
+    __omf.repo.git status --short --untracked-files
   end
 
-  functions -e omf.repo.git{,.stash}
   return 0
 end


pkg/omf/cli/omf.repo.pull.fish
```diff
@@ -8,11 +8,12 @@ function omf.repo.pull
   set -l repo_dir $argv[1]
 
   function omf.repo.git -V repo_dir
-    command git --git-dir "$repo_dir"/.git --work-tree "$repo_dir" $argv
+    command git -C "$repo_dir" $argv
```
derekstavis: Older version of OS X ships with Git < 1.8.5, which doesn't include `git -C` (see http://blogs.atlassian.com/2013/12/whats-new-git-1-8-5). Prefer keeping `--git-dir` with `--work-tree`.

pkg/omf/cli/omf.update.fish
```diff
@@ -15,14 +22,19 @@ function omf.update -a name
...
+  for path in {$OMF_PATH,$OMF_CONFIG}/$parent_path/$name
     not test -e "$path/.git"; and continue
 
-    omf.repo.pull $path; and set return_success
-
-    set -q return_success;
-      and omf.bundle.install $path/bundle
+    omf.repo.pull $path
+    switch $status
+    case 304
```
derekstavis: Shell ain't HTTP. I would prefer an approach where the function return one of:

+ `return -1`: Nothing to do
+ `return  0`: Updated
+ `return  1`: Error updating

oranja: That was my original idea. I only changed my mind because returning a negative value is not very common in *NIX as far as I know.

pkg/omf/cli/omf.repo.pull.fish
oranja: No problem, but let's -  big pretty please - decide what is the oldest `git` version OMF supports (sometime before OMFv1.0).
In #208 @bpinto said older OS X versions come with it's git 1.**8.7** and, you (Mr. @derekstavis :bowtie:) said that the installer uses shallow clones, and hence OMF requires even 1.**9**.
So let's get this confusion out of 

btw: I really hope, for the last bit of respect I have left for Apple, that upgrading Git to v2+ on those older OS X versions is a piece of cake and doesn't require Homebrew or anything like that.

bpinto: We should investigate this indeed.

pkg/omf/cli/omf.update.fish
bpinto: I don't know about negative values either, but I found the `304` at least funny. :P

pkg/omf/cli/omf.repo.pull.fish
bpinto: Apart from this change that would not support some versions of git (let's discuss this in a separate issue), I like the PR!

pkg/omf/cli/omf.update.fish
derekstavis: Yeah, negative numbers ain't common, but HTTP isn't too :stuck_out_tongue_closed_eyes:

pkg/omf/cli/omf.update.fish
```diff
@@ -15,14 +22,19 @@ function omf.update -a name
...
+      set update_status uptodate
+    case 0
+      set update_status success
+      omf.bundle.install $path/bundle
+    end
   end
 
-  set -q return_success; and __omf.update.success "$name"
+  eval __omf.update.$update_status "$name"
```
derekstavis: Is `eval` really needed??

pkg/omf/cli/omf.repo.pull.fish
```diff
@@ -20,15 +21,25 @@ function omf.repo.pull
...
-    or return $OMF_UNKNOWN_ERR
+  set initial_branch (omf.repo.git symbolic-ref -q --short HEAD)
+    or return (omf.repo.after_clean $OMF_UNKNOWN_ERR)
+  set initial_revision (omf.repo.git rev-parse -q --verify HEAD)
+    or return (omf.repo.after_clean $OMF_UNKNOWN_ERR)
+
+  # the refspec ensures that '$remote/master' gets updated
+  set -l refspec "refs/heads/master:refs/remotes/$remote/master"
+  omf.repo.git fetch --quiet $remote $refspec
+    or return (omf.repo.after_clean $OMF_UNKNOWN_ERR)
```
derekstavis: `or return $OMF_UNKNOWN_ERR`

pkg/omf/cli/omf.repo.pull.fish
```diff
@@ -20,15 +21,25 @@ function omf.repo.pull
...
+
+  # 304 = (HTTP) Not Modified
+  if test (omf.repo.git rev-list --count master...FETCH_HEAD) -eq 0
+    and return (omf.repo.after_clean 304)
```
derekstavis: `and return 2`

pkg/omf/cli/omf.repo.pull.fish
```diff
@@ -8,11 +8,12 @@ function omf.repo.pull
   set -l repo_dir $argv[1]
 
   function omf.repo.git -V repo_dir
-    command git --git-dir "$repo_dir"/.git --work-tree "$repo_dir" $argv
+    command git -C "$repo_dir" $argv
   end
 
-  function omf.repo.git.stash -V repo_dir
-    command git -C "$repo_dir" stash $argv
+  function omf.repo.after_clean -a return_code
```
derekstavis: I believe this is unecessary

pkg/omf/cli/omf.update.fish
oranja: It is really needed to call a function dynamically/indirectly, through a variable's value.
There is of course the alternative of using $update_status as an argument to a `switch..case` block.
I don't intend to deteriorate coding standards, so tell me which way you find more appropriate.

pkg/omf/cli/omf.repo.pull.fish
```diff
@@ -36,13 +47,11 @@ function omf.repo.pull
...
-  set -l refspec "refs/heads/master:refs/remotes/$remote/master"
-
-  # the refspec ensures that '$remote/master' gets updated
-  if not omf.repo.git pull --ff-only --quiet $remote $refspec
+  if not omf.repo.git merge --ff-only --quiet FETCH_HEAD
     omf.repo.git checkout $initial_branch
     omf.repo.git reset --hard $initial_revision
-    set -q stashed; and omf.repo.git.stash pop
+    set -q stashed; and omf.repo.git stash pop
+    return (omf.repo.after_clean $OMF_UNKNOWN_ERR)
```
derekstavis: `return 1`

pkg/omf/cli/omf.repo.pull.fish
oranja: So no cleanup of internal functions?

pkg/omf/cli/omf.update.fish
derekstavis: You already have a `switch..case` above. You can use it!

pkg/omf/cli/omf.repo.pull.fish
derekstavis: Although the cleanup function which return the parameter passed and cleans up functions is a nice idea, I found it somewhat cumbersome. But on the other side, the previous implementation had a bug, which removed the inner functions only when succeeded. I don't know a better way to solve this issue, but I would avoid masking the return code.

oranja:
There are numerous conflicts here that arise from me trying to stick to how things were previously, instead of doing things how they should be now and in the future (an approach I much prefer).
So instead of tip-toeing around, let's vote on a few topics (you can veto of course). Keep in mind that what we decide here we should try to maintain throughout the whole project. Consistency rocks.
1. Return codes - success is easy and is always 0. What other conventions do we have? How about different kinds of success (e.g. up-to-date)? What about negative numbers?
2. Cleanup of internal functions on return - yes or no? How is less important now.
3. Git version - I'm leaving it up to you to survey, bring to discussion and ultimately decide what is the minimum required for OMF.
4. Support for plugins under `$OMF_CONFIG` - yes or no? (if 'yes', why??)

bpinto:
> 1. Return codes - success is easy and is always 0. What other conventions do we have? How about different kinds of success (e.g. up-to-date)? What about negative numbers?

I would use positive numbers, similar to how `grep` does.
```
EXIT STATUS
     The grep utility exits with one of the following values:

     0     One or more lines were selected.
     1     No lines were selected.
     >1    An error occurred.
```

> 2. Cleanup of internal functions on return - yes or no? How is less important now.

I don't think this is important.

> 4. Support for plugins under $OMF_CONFIG - yes or no? (if 'yes', why??)

I don't understand the question we support plugins under $OMF_CONFIG so that people can use custom plugins that are not public available.

derekstavis:
>Cleanup of internal functions on return - yes or no? How is less important now.

The only change I would make to internal functions would be prefixing them with `__`, which would make them hidden to `functions -n`

oranja:
> we support plugins under $OMF_CONFIG so that people can use custom plugins that are not public available.

Let's clarify the question then: Any plugins installed by OMF are installed to `$OMF_PATH/{pkg,theme}`. If a package is under `$OMF_CONFIG`, it wasn't installed by OMF. Why should OMF try to update these custom packages? It's like asking Apple to update packages you installed with Homebrew. Or something like that... You get the idea.

bpinto:
Oh regarding the update... I agree, there is no need to try updating custom plugins.

oranja:
Took your comments into consideration and revised the code accordingly.
1. No evil `eval`.
2. Internal functions are prefixed with `__`. Not cleaned up on exit.
3. Simple `return`s with no delegation.
4. Return codes: 0 = success, 1 = error, 2 = up-to-date. So long 304 :cry: 
5. No more attempts to update packages under `$OMF_CONFIG`.
6. Still taking git 1.9.0+ as a minimal requirement (`git -C`). It's easy to lose this requirement here, but we can benefit from the newer git features in more places, so I'm leaving it for now.

pkg/omf/cli/omf.repo.pull.fish
```diff
@@ -7,55 +7,57 @@ function omf.repo.pull
...
+  set initial_revision (__omf.repo.git rev-parse -q --verify HEAD)
+    or return 1
+
+  # the refspec ensures that '$remote/master' gets updated
+  set -l refspec "refs/heads/master:refs/remotes/$remote/master"
+  __omf.repo.git fetch --quiet $remote $refspec
+    or return 1
+
+  if test (__omf.repo.git rev-list --count master...FETCH_HEAD) -eq 0
+    and return 2
```
derekstavis: `and` not needed :stuck_out_tongue_closed_eyes:

oranja: indeed.

oranja:
I pushed for testing. I think it still needs work. Do not merge or anything silly like that...

derekstavis:
The progress looks nice, but I would try checking for remote changes without fetching. This would also speed up the update process. Give a check [at this answer](http://stackoverflow.com/a/3278427/1429720) for more details.

oranja:
There might be a way, but `git remote update` eventually performs `git
fetch`

oranja:
There might be a way, but `git remote update` eventually performs `git
fetch`. See [here](http://stackoverflow.com/questions/2688251/what-is-the-difference-between-git-fetch-origin-and-git-remote-update-origin)

oranja:
So I think it's good now.
Will appreciate your help testing on different platforms + any other feedback.

This PR does provide a more accurate output for updates, but sadly, it still doesn't help much with speed and bandwidth. So far I couldn't find a way to check for updates without fetching. I have two other ideas on how to mitigate this problem:

1. Concurrency. I am willing to bet that the bottleneck is the network. I suppose that it's not related to network "speed" as much as it is related to network overhead (TCP handshake, encryption and so forth) . Hence, I suppose that splitting the update job into a few threads will do great to the process. (How exactly to pull this off, I'm not sure yet :bowtie:). For safety measures, I suggest only doing the fetches in parallel and then do the checks and merges sequentially. [:+1: speed; :see_no_evil: bandwidth; :question: complexity]
2. Keep an index. This means that some server polls the different packages in the database every once in a while and keeps the timestamp of the last commit to the repository. `omf update`s, in turn, will save the most recent update timestamp and pull only to the outdated repositories.  [:+1: bandwith, speed; :-1: complexity]

Perhaps it's worth discussing this in a separate issue. I will try to work out the details if I have the chance, but I have other WIP that I don't want to leave hanging.

bobthecow:
Looks good to me.

PR #221, opened 2016-01-02
git_ahead: Rewrite code, fix ahead being reported as behind

oranja:
I'm not sure that this is a great change, but perhaps taking out the `awk` script and doing usual Fish checks is a more reasonable approach?

### DIFF
--- a/lib/git/git_ahead.fish
+++ b/lib/git/git_ahead.fish
@@ -1,15 +1,22 @@
 function git_ahead -a ahead behind diverged none
-  git_is_repo; and begin
-    test -z "$ahead"; and set ahead "+"
-    test -z "$behind"; and set behind "-"
-    test -z "$diverged"; and set diverged "±"
-    test -z "$none"; and set none ""
-    command git rev-list --left-right "@{upstream}...HEAD" ^/dev/null \
-    | awk "/>/ {a += 1} /</ {b += 1} \
-      {if (a > 0) nextfile} END \
-      {if (a > 0 && b > 0) print \"$diverged\"; \
-      else if (a > 0) print \"$ahead\"; \
-      else if (b > 0) print \"$behind\";
-      else printf \"$none\"}"
+  not git_is_repo; and return
+
+  set -l commit_count (command git rev-list --count --left-right "@{upstream}...HEAD" ^/dev/null)
+
+  switch "$commit_count"
+  case ""
+    # no upstream
+  case "0"\t"0"
+    test -z "$ahead"; and echo "$ahead"
+      or echo ""
+  case "*"\t"0"
+    test -z "$behind"; and echo "$behind"
+      or echo "-"
+  case "0"\t"*"
+    test -z "$ahead"; and echo "$ahead"
+      or echo "+"
+  case "*"
+    test -z "$diverged"; and echo "$diverged"
+      or echo "±"
   end
 end


bobthecow:
bobthefish uses something a lot more like that awk script because it is _way_ faster.

oranja:
If that is the case, this can be closed :+1:

bpinto:
Regarding any git function, I think the best thing to do is to use the code that exists in the git repository for aurocomplete. Is that what you did @bobthecow?

oranja:
[git-prompt.sh] implements the "git-ahead" function in two variants:
1. If the version of Git is recent enough to support `rev-list --count`, then `--count` is used and then a simple `switch..case` logic is used to determine if ahead/behind/diverged/none. In verbose mode, [bobthefish] skips the `--count` option but uses the same `switch..case` logic. I can take the `cut`s and the `if..else`s out and favor the `switch..case` logic too.
2. Legacy mode means `rev-list --count` cannot be used. Instead, commits are summed with yet another loop of `switch..case`s. [bobthefish] uses `awk` to get the same result, with the second step of determining ahead/behind/diverged/none also incorporated into the same `awk` script.

I won't be immensely surprised if Fish does not evaluate the `switch..case`s as quickly as Bash or `awk`, but if it's a very small sacrifice, I personally prefer to use Fish tools.
@bobthecow in your benchmarks, what exactly was the performance of `awk` compared to?

This is the relevant excerpt from [git/contrib/completion/git-prompt.sh][git-prompt.sh]:
```bash
# Find how many commits we are ahead/behind our upstream
if [[ -z "$legacy" ]]; then
	count="$(git rev-list --count --left-right \
			"$upstream"...HEAD 2>/dev/null)"
else
	# produce equivalent output to --count for older versions of git
	local commits
	if commits="$(git rev-list --left-right "$upstream"...HEAD 2>/dev/null)"
	then
		local commit behind=0 ahead=0
		for commit in $commits
		do
			case "$commit" in
			"<"*) ((behind++)) ;;
			*)    ((ahead++))  ;;
			esac
		done
		count="$behind	$ahead"
	else
		count=""
	fi
fi

# calculate the result
case "$count" in
"") # no upstream
	p="" ;;
"0	0") # equal to upstream
	p="=" ;;
"0	"*) # ahead of upstream
	p=">" ;;
*"	0") # behind upstream
	p="<" ;;
*)	    # diverged from upstream
	p="<>" ;;
esac
```

[bobthefish]: https://github.com/oh-my-fish/theme-bobthefish/blob/master/fish_prompt.fish
[git-prompt.sh]: https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh

bpinto:
I would keep the fastest code possiblr as it is a code that is executed every time (prompt style).

oranja:
So I ran some benchmark tests comparing 5 different methods for this small utility function.
Following are the results, averaged over {3 trials x 4 repo statuses} for each method. Each trial with 10000 calls to `git_ahead`:
git_ahead_chart
1. `--count` + `switch..case`
real: 19.8575 :crown:
user: 8.945 :crown:
sys: 11.4675 :crown:
user+sys: 20.4125 :crown:
2. `--count` + `if..else`
real: 37.3191666667
user: 13.7458333333
sys: 27.9816666667
user+sys: 41.7275
3. no `--count` + `awk`
\* awk script taken from **lib/git/git_ahead.fish** :tropical_fish:
\* wrong reports excluded from average
real: 20.0111111111
user: 12.9644444444
sys: 16.7977777778
user+sys: 29.7622222222
4. no `--count` + `if..else`
real: 43.355 :hankey:
user: 17.3575 :hankey:
sys: 29.1716666667 :hankey:
user+sys: 46.5291666667 :hankey:
5. no `--count` + `switch`
real: 42.7108333333
user: 16.9033333333
sys: 29.0391666667
user+sys: 45.9425

I should mention that currently the `git_ahead` function contains yet another bug. **diverged** branches are identified as **ahead** instead. Since I am currently using _bobthefish_, which has its internal implementation of `git_ahead`, I got accurate results on the prompt, but calling `git_ahead` gave me the wrong results. Now at least we can rest assured that something good has come out of all this messing around with such a simple function. :stuck_out_tongue_winking_eye:
On the one hand, it might be a little unfair to run the bogus version of the `awk` script. On the other hand, it shows that we can get comparable results to what we have now, but in a bug-free (I hope :sweat_smile:) script. If anyone cares to reproduce with the correct `awk` script, they're welcome.

Having all this said and done, we still need to consider the git compatibility issue and decide which version of git should be the minimum requirement of OMF. If we decide on anything older than 1.9.0 (I think I'm rounding up, but there are many useful features in 1.9.0 -- see #208).

You can find the `git_ahead` variants, as well as the benchmark runner and results at [oranja/omf-benchmarks](oranja/omf-benchmarks/tree/master/git_ahead).

derekstavis:
>I should mention that currently the git_ahead function contains yet another bug. diverged branches are identified as ahead instead.

Yes. This is true. And I have fixed it here: https://github.com/oh-my-fish/plugin-vcs/blob/master/functions/git/vcs.status.fish

Give a look at how I've implemented things on vcs plugin. The idea behind it is to be used by all prompts, as vcs plugin transparently handles all kinds of repositories (git, hg and svn).

oranja:
1. I see that you, too, use the `--count` feature, which is only available from git 1.8.5.3. If we can assume at least 1.8.7, no complains. Otherwise, a potential problem. I hate to do it (I don't), but here I'm pointing again at #220, where we decided to reject the use of `git -C` because it's only available from git 1.8.5.
2. It was my first idea to use the straightforward `if..else`, but as you can see above, `--count`+`switch..case` are superior to `--count`+`if..else`.
And if we do drop the `--count`, it seems better to use the `awk` trick.
(performance-wise)

derekstavis:
Using awk is even worse on OS X, where forks are very costly. I'm +1 for using fish builtins. About `--count` I think it's time to start supporting Git 1.9+.

oranja:
This PR is now officially a bugfix.

Unofficially, it's me trying to make the code just a little bit less cryptic, without hurting performance.

lib/git/git_ahead.fish
```diff
@@ -1,15 +1,23 @@
...
+  test -z "$diverged"; and set diverged "±"
+  test -z "$none"; and set none ""
+
+  set -l commit_count (command git rev-list --count --left-right "@{upstream}...HEAD" ^/dev/null)
+
+  switch "$commit_count"
+  case ""
+      # no upstream
+  case "0"\t"0"
+      echo "$none"
```
bpinto: I wonder if it would be faster to do test; and echo instead of doing it previously and assign it to a variable.

oranja:
Trying to shave a few more milliseconds. Thanks for the suggestion @bpinto

derekstavis:
It looks good to me!

PR #224, opened 2016-01-03
Update package: fasd

simnalamburt:
1. ~~https://github.com/simnalamburt/pkg-cgitc~~
2. https://github.com/simnalamburt/pkg-fasd
3. ~~https://github.com/simnalamburt/pkg-sensible~~

--------

#### Note for fasd
Existing fasd plugin doesn't even work properly.

- No fasd_cd
- No default aliases (f, a, s, d, z, zz, ...)

So I forked it and totally rewrote it. Please consider it as different plugin.

### DIFF
--- a/db/pkg/fasd
+++ b/db/pkg/fasd
@@ -1,1 +1,1 @@
-https://github.com/wa/pkg-fasd
+https://github.com/oh-my-fish/plugin-fasd


derekstavis:
Hey @simnalamburt, thanks for your work! We don't usually accept packages with aliases (or abbreviations), as they are very personal. For example: The `cgit` package overrides some of my personal aliases. Also does `fasd` aliases (`f` for me opens OS X Finder). We recommend you keeping aliases packages into your user namespace and installing by URL (so you have a bundle which reflects your installed modules).

About `fasd` package, would you like us to allocate a repository inside Oh My Fish organization for you and add you into organization as a maintainer? This way packages can be neatly searched via organization repository list in Github website.

simnalamburt:
@derekstavis Ok I push forced and removed the aliases from the PR.
And about `fasd`, yes! I'm very honored to be in Oh My Fish organization.

simnalamburt:
And about `fasd`'s default aliases, I'm quite sure that uses will wants those aliases.

`fasd` intentionally aliases those characters and almost no one use the fasd with command `$ fasd arg` or `$ fasd_cd arg` directly.
https://github.com/clvv/fasd/blob/master/fasd#L82

derekstavis:
If it's to mimic the way fasd works it's a good argument. :100:

I have allocated a repository for you inside Oh My Fish organization and added you as a collaborator. You can push your work [there](https://github.com/oh-my-fish/plugin-fasd) and modify the PR to reflect the address change, then I merge it. Welcome to the organization!! :+1: :+1:

derekstavis:
Awesome! Thanks @simnalamburt

PR #226, opened 2016-01-04
Remove update instructions

bpinto:
Oh My Fish has been stable for many months, we no longer need the update
instructions on the main README.


derekstavis:
I think we're good to remove it :+1:

PR #228, opened 2016-01-04
Remove test-generate-themes-doc

derekstavis:
This tests isn't helping us with the process of merging new themes.


bpinto:
Why not? I think its valuable

derekstavis:
I see it as another barrier for adding themes to database. We could generate this theme documentation after merging new themes to database.

PR #230, opened 2016-01-10
Remove PATH save/restore code

### LINKED ISSUE #186: Changes to PATH and new shells
Inside a fish shell, any changes to the `PATH` are not reflected on any new shells. Eg:
```
$ set -x PATH /foo/bar $PATH
$ echo $PATH
  /foo/bar <rest of path>
$ fish
$ echo $PATH
  <rest of path>
```

Inside `init.fish` there is a block near the top with the following:
```
# Save PATH before oh my fish for reseting the PATH when we reload OMF.
if set -q ORIGINAL_PATH
  set PATH $ORIGINAL_PATH
else
  set -gx ORIGINAL_PATH $PATH
end
```

It seems that the intended functionality here is to prevent OMF scripts from duplicating additions to the `PATH` when reloading. The (likely) unintended functionality is that it blocks any edits to the `PATH` regardless of them originating from OMF.

For now I have added `set -e ORIGINAL_PATH` in my own `init.fish` script to prevent this behavior, but I'm thinking there should be a better way to accommodate the desired effect without sacrificing other edits.

derekstavis:
Fixes #186 by removing the PATH save/restore code from init.fish.

There isn't a single occurence where Oh My Fish code changes the
value of PATH variable.

Plugin code which changes the PATH variable should be aware of
the side-effects and manage PATHs correctly, avoiding duplication.

### DIFF
--- a/init.fish
+++ b/init.fish
@@ -14,19 +14,10 @@
 #   + Source $OMF_CONFIG/init.fish
 #
 # ENV
-#   OSTYPE        Operating system.
-#   ORIGINAL_PATH Original $PATH preseved across Oh My Fish reloads.
 #   OMF_PATH      ~/.local/share/omf by default.
 #   OMF_IGNORE    List of packages to ignore.
 #   OMF_CONFIG    ~/.config/omf by default.
 #   OMF_VERSION   Oh My Fish! version
-
-# Save PATH before oh my fish for reseting the PATH when we reload OMF.
-if set -q ORIGINAL_PATH
-  set PATH $ORIGINAL_PATH
-else
-  set -gx ORIGINAL_PATH $PATH
-end
 
 # Set OMF_CONFIG if not set.
 if not set -q OMF_CONFIG


oranja:
I think it's a good idea to get rid of this overriding, especially if it has no use.

In fact I'd also love to see these lines changed:
```fish
# Save the head of function path and autoload core functions
set -l user_function_path $fish_function_path[1]
set fish_function_path[1] $OMF_PATH/lib
```

If the purpose is to give precedence to all OMF functions, perhaps it's enough to prepend OMF's path to the beginning of the list?
If the purpose is to override specific functions, then which? and why? The answers should be noted somewhere.
In the same manner, if the objective is to hide specific functions, which? why?

derekstavis:
>If the purpose is to give precedence to all OMF functions

The original idea of this ugly code is to give precedence to user function path, which is always the first item when the shell is cleanly-initialized. Please be aware of the `autoload $user_function_path`:

```fish
set -l user_function_path $fish_function_path[1]
set fish_function_path[1] $OMF_PATH/lib

# Autoload util functions
autoload $OMF_PATH/lib $OMF_PATH/lib/git

...

autoload $user_function_path
```

oranja:
I am aware, but `autoload $user_function_path` comes only after all plugins are loaded, which in effect hides custom functions from plugins.
This prevents plugins from doing Fish-style customizations (e.g. define a function called 'myplugin-user-config' that sets variable `myplugin-urls` with a list of rss feeds to track).
As long as OMF doesn't define some other suitable place for such functions, I find the hiding behavior uncalled for.

Plus, eventually `autoload` works by getting `$user_function_path` back into `$fish_function_path`, right?
So I think it's worth to check if we can demote `$user_function_path` in one shot, without hiding it from plugins.

derekstavis:
Merged as 309fa3b

PR #231, opened 2016-01-12
registry: add rafaelrinaldi/theme-pure

oranja:
Following oh-my-fish/#227, @rafaelrinaldi @edouard-lopez joined forces to make the upstream repository compatible with OMF, including a "theme-pure" alias.
**[pure]** can now join the family of OMF's registered themes. :tada: 

[pure]: rafaelrinaldi/pure

PR #233, opened 2016-01-13
Update theme docs (Themes.md)

oranja:
I missed an update to [pure]'s Readme.md in the last PR (#231). :flushed:
Should fix the failing build.

[pure]: /rafaelrinaldi/pure


derekstavis:
Oh, so that was the problem! Thanks for fixing :+1:

PR #241, opened 2016-01-20
Add chain theme to db

coderstephen:
I made a new theme! Add it to the database?

Hopefully the theme isn't too similar to other themes -- I tried to distinguish it a bit. Check it out here: https://github.com/oh-my-fish/theme-chain.

### DIFF
--- a/db/themes/chain
+++ b/db/themes/chain
@@ -1,0 +1,1 @@
+https://github.com/oh-my-fish/theme-chain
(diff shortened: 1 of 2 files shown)


derekstavis:
Don't forget to run `tools/generate-themes-doc.fish` :)

coderstephen:
Whoops, forgot! :ghost: Fixed.

PR #247, opened 2016-01-25
init: remove dupe in fish_function_path

derekstavis:
Self explaining.


init.fish
```diff
@@ -33,7 +33,7 @@ set -l user_function_path $fish_function_path[1]
 set fish_function_path[1] $OMF_PATH/lib
```
oranja: This makes me cringe every time I see it. :confounded: 
Wouldn't it be more uniform and make more sense to:

1. push the first entry to the end instead of erasing it and putting it back later?
2. `autoload` both **lib** and **lib/git** in the same way?

PR #262, opened 2016-03-04
autoload: refactor for conformance and performance

derekstavis:
Refactor autoload: Split the big function into two smaller ones,
doing only option parsing at main function.

The algorithm is also rewritten, now in two steps for both path
inclusion an exclusion functions: 1) use auxiliary lists to store
valid function and completion paths, 2) bulk insert or remove just
once in the variable.

Now also respects path insertion policy, keeping user function
path always in front of other paths, thus allowing precedence of
user functions.

### DIFF
--- a/lib/autoload.fish
+++ b/lib/autoload.fish
@@ -1,58 +1,52 @@
-# SYNOPSIS
-#   autoload <path>...
-#   autoload -e <path>...
-#
-# OVERVIEW
-#   Manipulate autoloading path components.
-#
-#   If called without options, the paths passed as arguments are added to
-#   $fish_function_path. All paths ending with `completions` are correctly
-#   added to $fish_complete_path. Returns 0 if one or more paths exist. If all
-#   paths are missing, returns != 0.
-#
-#   When called with -e, the paths passed as arguments are removed from
-#   $fish_function_path. All arguments ending with `completions` are correctly
-#   removed from $fish_complete_path. Returns 0 if one or more paths erased. If
-#   no paths were erased, returns != 0.
-
-function autoload -d "Manipulate autoloading path components"
-  set -l paths $argv
-
+function autoload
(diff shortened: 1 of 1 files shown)


lib/autoload.fish
```diff
@@ -1,58 +1,53 @@
...
-    end
-  end
-
-  set -q return_success
-end
+      contains -- $path $fish_function_path
+        or set function_path $function_path $path
+    end;
+  end;
+  test "$function_path"
```
oranja: Wouldn't a more explicit `test -n` be more appropriate here?
(Same goes for line#34)

lib/autoload.fish
```diff
@@ -1,58 +1,53 @@
...
-#   paths are missing, returns != 0.
-#
-#   When called with -e, the paths passed as arguments are removed from
-#   $fish_function_path. All arguments ending with `completions` are correctly
-#   removed from $fish_complete_path. Returns 0 if one or more paths erased. If
-#   no paths were erased, returns != 0.
-
-function autoload -d "Manipulate autoloading path components"
+function autoload
   set -l paths $argv
```
oranja: Not used anywhere anymore

derekstavis: It's used below!

lib/autoload.fish
derekstavis: Yeah, it's also possible

lib/autoload.fish
oranja: Just to be clear, I'm talking about `$paths`, and nope... :no_good:

derekstavis: Oh gosh, I was looking at the removed stuff :laughing:

derekstavis:
Done! Rebased to current master!

oranja:
:+1: :100:

Issue #252, opened 2016-01-25
omf theme completions show available but not installed themes


derekstavis:

oranja:
Based on @decors suggestion in #251 and current [code], I suggest:
```fish
complete -c omf -f -n "__fish_seen_subcommand_from t theme" -a (omf.packages.list --installed --theme)
```

I really don't know when I will get the chance to submit this, so anyone is welcome to suggest this (or something even better) in a PR.

[code]: https://github.com/oh-my-fish/oh-my-fish/blob/master/pkg/omf/completions/omf.fish

coderstephen:
Fixed by #263.

Issue #251, opened 2016-01-25
omf install completions doesn't show themes


derekstavis:
Currently install completions does not show available themes.

decors:
how about like this?

```
function __omf_install_completion
  for theme in (omf.packages.list --available --theme)
    printf "%s\ttheme\n" $theme
  end
  for plugins in (omf.packages.list --available --plugins)
    printf "%s\tplugins\n" $plugins
  end
end

complete -c omf -f -n "__fish_seen_subcommand_from i install" -a "(__omf_install_completion)"
```

oranja:
This seems to do  a great start, but can you explain to someone who never did completions before (me), why there's a need to print "theme" after each theme name, and "plugins" after each plugin name?

In other words, why can't this simply be:
```fish
complete -c omf -f -n "__fish_seen_subcommand_from i install" -a "(omf.packages.list --available)"
```
?

decors:
@oranja Thanks

My code has some errors

```
function __omf_install_completion
  for t in (omf.packages.list --available --theme)
    printf "%s\ttheme\n" $t
  end
  for p in (omf.packages.list --available --plugins)
    printf "%s\tplugin\n" $p
  end
end

complete -c omf -f -n "__fish_seen_subcommand_from i install" -a "(__omf_install_completion)"
```

```
> omf install 
agnoster             (theme)
agnoster-mercurial  (plugin)
ansible                     
apt                 (plugin)
artisan                     
balias              (plugin)
bang-bang           (plugin)
```

oranja:
Oh! Now I see why you put "theme"/"plugin" after each option. That actually gives a great user experience in my opinion. :)

I don't see any error in your code now. Run `complete -e -c omf -f -n "__fish_seen_subcommand_from i install"` before your code to erase all current completions for `omf install`. You'll see that your code produces the desired results.

I noticed, however, that breaking down the completion into smaller commands results in a more responsive shell. Can someone else confirm that It's not my imagination?

```fish
complete -e -c omf -f -n "__fish_seen_subcommand_from i install"
```
```fish
for t in (omf.packages.list --available --theme)
  complete -c omf -f -n "__fish_seen_subcommand_from i install" -a "(printf '%s\ttheme' $t)"
end
for p in (omf.packages.list --available --package)
  complete -c omf -f -n "__fish_seen_subcommand_from i install" -a "(printf '%s\tpackage' $p)"
end
```

coderstephen:
[...]

PR #263, opened 2016-03-05
Improve OMF CLI completions

oranja:
This is in response to issues #251 and #252.
Changes:
+ Completions for `remove`
+ `cd` and `install` show both plugins and themes
+ `theme` shows only installed themes

It might be a bit slower to initialize, but provides more responsive autosuggestions (according to my short experiments).


derekstavis:
Cool :+1:

bobthecow:
Thanks! :+1: :+1:

PR #265, opened 2016-03-06
Update plugin URLs and remove broken plugin entries

coderstephen:
A number of plugins in the package database have moved repository locations, or have simply been deleted. This PR updates the URLs of any plugins to their new GitHub locations, and removes entries to plugins that are no longer available.

### DIFF
--- a/db/pkg/ansible
+++ b/db/pkg/ansible
@@ -1,1 +1,1 @@
-https://github.com/wa/pkg-ansible
+https://github.com/pkg-gretel/pkg-ansible
--- a/db/pkg/copy
+++ b/db/pkg/copy
@@ -1,1 +1,0 @@
-https://github.com/wa/pkg-copy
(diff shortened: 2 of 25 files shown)


db/themes/pure
```diff
@@ -1 +1 @@
-https://github.com/rafaelrinaldi/theme-pure
+https://github.com/rafaelrinaldi/pure
```
oranja: We still need the `theme-` prefix for install, IIRC

bobthecow: Right. The redirect there was a favor to OMF. Let's hold off on breaking our packages until we support prefixless repo names :)

derekstavis: +1 for keeping the `theme-` prefix. Unfortunately it's still needed.

coderstephen:
Sorry about the delay. Just put the `theme-` prefix back.

derekstavis:
I think this can be merged. Rebase?

coderstephen:
Literally just did. :+1:

derekstavis:
That was fast :)

PR #268, opened 2016-03-11
Allow selective updates

oranja:
Until now, `omf update` was always "all-in".
This proposal keeps the default behavior, but also let's the user choose to update only the core framework and/or some packages.

##### Usage examples:
- `omf update`: update the framework and all installed packages (default behavior)
- `omf update omf`: update the core framework only
- `omf update agnoster`: only update the agnoster theme
- `omf update balias omf agnoster`: update the core framework, the balias plugin and the agnoster theme. The core framework is always updated first, if included.

PR #269, opened 2016-03-12
Parallel fetch on `omf update`

oranja:
The current `omf update` process is somewhat slow, and becomes even more slow with each new installed package. The immediate suspects are the sequential `git fetch`s that are done for each and every installed package.

Based on the following assumptions:
- A single `git fetch` instance is not too CPU or Disk I/O intensive to exclude other instances from running in parallel.
- Each `git fetch` instance is slowed down a lot more by network overhead (DNS resolve, TCP handshake, session encryption) than by CPU and I/O operations.
- Each `git fetch` instance in the `omf update` process works on a different and independent directory on the file-system.
- It is reasonable to defer status reports for each package -- until the merge phase, in favor of a more simple implementation.

The `fetch` phase seems like an excellent candidate for concurrent execution.
Running several `git fetch` instances in parallel might benefit from concurrent execution on multiple threads/cores, and most importantly from the option to keep working while a single instance opens a connection to github's servers
Eventually these advantages reduce the total time required to perform an `omf update`.

_Performance_:
1. Cursory benchmarks, tested by updating core and 40 installed plugins and themes, all already up to date.
   `time -p fish -ic "omf.cli.update"`
   - Without the proposed changes (HEAD at fb004f8)
     real **40.65**
     user **4.02**
     sys **5.25**
     (You can see that this process is not CPU-bound. Something else wastes most of the time)
   - With the proposed changes (branch [parallel-fetch](https://github.com/oranja/oh-my-fish/commits/parallel-fetch))
     real **10.41**
     user **11.63**
     sys **8.41**
     (You can see that multiple cores are actually used: `real` < `user`+`sys` -by a large margin.)
   
   Overall, the process is 4 times faster on my machine with the proposed changes.
2. Updating a single package _vs._ updating two packages.
   - With the proposed changes (branch [parallel-fetch](https://github.com/oranja/oh-my-fish/commits/parallel-fetch), including the changes from #268)
     - `time -p fish -ic "omf.cli.update balias"`
       real **1.39**
       user **0.51**
       sys **0.51**
     - `time -p fish -ic "omf.cli.update balias agnoster"`
       real **1.46**
       user **0.72**
       sys **0.74**
   
   Adding another package to the queue only increased the total run-time by 5%.
   - Without the proposed changes (HEAD at fb004f8)
     - `time -p fish -ic "omf.packages.update balias"`, without the proposed changes.
       real **1.19**
       user **0.25**
       sys **0.23**
     - `time -p fish -ic "omf.packages.update balias; omf.packages.update agnoster"`, without the proposed changes.
       real **2.21**
       user **0.38**
       sys **0.35**
   
   A simple check without the proposed changes showed an 85% increase (1.19 real for 1 vs. 2.21 real for 2).
3. There is an evident drawback to this change, however. The proposed `update` procedure relies on busy-wait loops which loads the system even more than the actual operation.

_Testing_: This has been tested on my 5-year-old dual-core i7 laptop, running Fedora 23 x64 and fish v2.2.0. Help testing this for functionality and performance on OS X and any other platform will be very welcome.

_Note_: This PR relies heavily on PRs #220 and to a little extent on #268.
Since they both received no objections so far, I allowed myself to assume that they will be accepted soon, and use them as the base for this PR.

_Possible future improvement_: Provide more detailed progress information in the fetch phase.
It's easy to get simple progress information by tracking `$argv` while it's getting shorter. I chose to skip this enhancement at this time because of personal time constraints, and because I prefer the main functionality changes to get proper attention first.

### DIFF
--- a/pkg/omf/functions/cli/omf.cli.update.fish
+++ b/pkg/omf/functions/cli/omf.cli.update.fish
@@ -30,9 +30,14 @@ function omf.cli.update
   omf.index.update
     or return 1
 
-  for package in $packages
-    omf.packages.update $package
+  if omf.packages.check-updates $packages
+    for package in $packages
+      omf.packages.update $package
+    end
+  else
+    echo (omf::em)"All packages are up to date."(omf::off)
   end
+
 
   # Opt-in flag for testing
   if set -q OMF_AUTO_RELOAD
--- a/pkg/omf/functions/packages/omf.packages.check-updates.fish
+++ b/pkg/omf/functions/packages/omf.packages.check-updates.fish
@@ -2,7 +2,7 @@ function omf.packages.check-updates
   set -l max_workers 8
   set -l fetch_cmd "omf.repo.fetch"
   set -l progress_indicator \u25cf
-  set -l error_log (mktemp -t omf_check-updates.log)
+  set -l error_log (mktemp -t omf_check-updates.XXXX.log)
 
   # Supress Git's attempt to find missing repositories with different credentials.
   # Git > 2.3
@@ -24,7 +24,7 @@ function omf.packages.check-updates
 
         # Error reporting here would be ugly because it clutters the display.
         # Save to log file instead.
-        fish -c "$fetch_cmd $package_path" ^$error_log &
+        fish -c "$fetch_cmd $package_path" ^^$error_log &
 
         printf "$progress_indicator"
       else
@@ -41,6 +41,6 @@ function omf.packages.check-updates
   end
   echo " done!"
 
-  test (wc -l <$error_log) -gt 0
+  test (wc -l < $error_log) -gt 0
     and echo (omf::err)"Please see "(omf::off)"$error_log"(omf::err)" for errors encountered while getting updates."(omf::off)
 end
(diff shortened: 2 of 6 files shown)


oranja:
I'm putting this on hold until I investigate a report on hangs due to this change

oranja:
I made some changes and I think it should be working good now also in some edge cases.
Would love me some feedback on this one before it's merged.

oranja:
(Also) More than a year late, but better late than never, they say.

As far as I could see, only small changes were required to make this work again.
Extra eyes making sure that I didn't not miss any important change over this year (either in _fish_ or _omf_) can still help to make sure.

derekstavis:
It's working as intended for me!! The only thing I would change is the loading indicator.

Maybe a progress bar that is replaced by the success message in the end would be better? 😛

oranja:
As per your suggestion, @derekstavis, I implemented some kind of progress bar, but it's not the prettiest code and adds a lot of LoC, relatively, to this PR.

As I said before, I think it's better to merge the core of this PR (first 3 commits) first to see that the main functionality works as expected for everyone on 'dev', then it can be open to anyone to play with eye-candy.

Anyway, review and thoughts are most welcome.

faho:
This has gone stale. Is anyone interested in resolving the conflicts or should it be closed?

PR #276, opened 2016-03-15
lib: deprecate `refresh` function

derekstavis:
With great power comes great responsibility. This function is
somewhat low level, and is harmful to shell's user, as it
completely screws up with job control (subprocesses).

This PR officially marks `refresh` as deprecated, moving the
deprecated functionality to omf plugin `compat` quarantine
directory.


bobthecow:
How does the `</dev/tty` removal affect this?

derekstavis:
In my tests it had no effect, neither on OS X or Linux. I believe `</dev/tty` was added in a try to solve issues with Travis.

pkg/omf/functions/compat/refresh.fish
```diff
@@ -0,0 +1,14 @@
+function refresh -d "(deprecated) Refresh fish session by replacing current process"
+  printf 'warning: function %srefresh%s is deprecated and will be removed soon.\n %s\n' \
+  (set_color -u) (set_color normal) \
+  (status -t)[5] >&2
```
bobthecow: it's not clear to me what `(status -t)[5]` is intended to return.

derekstavis: It gets the fifth line of `status -t` :trollface:

derekstavis: I could split the stack trace and the index expansion to make it clearer what `status -t` output is, or I could use `status --print-stack-trace`.

bobthecow: Right. But what is it supposed to be?

derekstavis: It prints where the function was called.

bobthecow: I can't get it to say anything but standard in.

derekstavis:
```
function foo
  refresh
end
funcsave foo
foo
```

derekstavis: This is done so you can discover the offending file

bobthecow: it always literally just says ` 	called on standard input` for me.

bobthecow: with some extra spaces at the start.

derekstavis: for me this is the output of calling function `foo`

```
⋊ > ~/.l/s/omf on deprecate-refresh ● foo
warning: function refresh is deprecated and will be removed soon.
 	called on line 2 of file ~/.config/fish/functions/foo.fish
```

bobthecow:
```
warning: function refresh is deprecated and will be removed soon.
 	called on standard input
```

derekstavis: have you saved using `funcsave foo`? :)

bobthecow: kk. if i call it from a function defined in a file, it works. let's `grep -v` to keep it from talking about standard input and call it a day?

derekstavis: I don't got your point. You mean hiding the message when it's called in standard input?

bobthecow: i mean no matter how i call it, it only says "called on standard input" for me. even when that's clearly not the case. i feel like, given this, and given the inability to do anything with that information even if it was true, it makes the most sense to suppress "called on standard input". it's just not useful information :)

pkg/omf/functions/compat/refresh.fish
```diff
@@ -0,0 +1,14 @@
...
+
+  set -q CI
+    and return 0
+
+  type -q omf.core.reload
+    and omf.core.reload
+
```
oranja: A comment here that `omf.core.reload` calls `exec` and that the lines below are only fallback will be helpful IMO.

bobthecow: :+1:

derekstavis:
Here we go! Fixed the pointed issues and also rebased to current master! :tada:

bobthecow:
:+1:

coderstephen:
:+1:

oranja:
Looks great to me :+1:

bobthecow:
Maybe add the replacement to the error message?

scorphus:
:+1:

PR #280, opened 2016-03-19
`omf doctor` checks for fish & git versions

oranja:
Minimum *fish* version: *2.2.0*
* OMF uses `function -V` (`--inherit-variable`) in omf.repo

Minimum *git* version: *1.9.5*
* OMF uses `git -C` in omf.repo
* OMF uses `git --depth 1` in installer
* OMF intends to use `git --depth 1` for package install

### DIFF
--- a/pkg/omf/functions/util/omf.check.version.fish
+++ b/pkg/omf/functions/util/omf.check.version.fish
@@ -1,0 +1,3 @@
+function omf.check.version -a min_version -a current_version
+  test (echo "$min_version"\n"$current_version" | tr '.' ' ' | sort -n | head -n1) = (echo "$min_version" | tr '.' ' ')
+end
--- a/pkg/omf/functions/core/omf.doctor.fish
+++ b/pkg/omf/functions/core/omf.doctor.fish
@@ -3,7 +3,6 @@ function __omf.doctor.theme
     echo (omf::err)"Warning: "(omf::off)(omf::em)"fish_prompt.fish"(omf::off)" is overridden."
     echo (omf::em)"  fish_config"(omf::off)" command persists the prompt to "(omf::em)"~/.config/fish/functions/fish_prompt.fish"(omf::off)
     echo "  That file takes precedence over Oh My Fish's themes. Remove the file to fix it:"
-
     echo (omf::em)"  rm ~/.config/fish/functions/fish_prompt.fish"(omf::off)
     echo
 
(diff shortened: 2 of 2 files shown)


pkg/omf/functions/util/omf.check.version.fish
```diff
@@ -0,0 +1,3 @@
+function omf.check.version -a min_version -a current_version
+  test (echo "$min_version"\n"$current_version" | sort -V | head -n1) = "$min_version"
```
bobthecow: `sort -V` isn't portable. BSD and OS X sort doesn't have it.

bobthecow:
```fish
 function omf.check.version -a min_version -a current_version
  set -l IFS .
  echo $min_version | read -la min_chunks
  echo $current_version | read -la current_chunks
  set -l current_chunks_count (count $current_chunks)
  for i in (seq (count $min_chunks))
    set -l chunk 0
    math "$i <= $current_chunks_count" >/dev/null ^&1
      and set chunk $current_chunks[$i]
    math "$min_chunks[$i] <= $chunk" >/dev/null ^&1
      or return 1
  end
end
```

bobthecow: the current chunk count thing was because i couldn't figure out a way to make the array stuff happy when comparing, for example, min version 1.2.3 with current version 1.2.

any better ideas?

oranja: Oh... Right. So I need to add something like
```fish
uname -a | grep -qEv "os x|bsd"; or echo "goddammit" >&2
```

j/k. alternate version coming in a second

bobthecow: _my_ version comes up with the right answer when checking `omf.check.version 3.11.0 3.11` … but i guess that's an edge case that'll probably never be needed :P

bobthecow:
:+1:

pkg/omf/functions/util/omf.check.version.fish
oranja: Actually I was just checking my changes when you posted your version, so I only saw it when the PR was already updated. Otherwise I might have used your code.
On the upside, after (/ if) my edge-case-failing code is merged, you can override it with yours and get proper credits.

bobthecow: :) I'm not worried about credit.

bobthecow:
Everyone OK with these minimum versions?

scorphus:
Git 1.9.5 is more than 1 year old. Latest Fish release, 2.2, is 9.5 months old. So, yeah :+1: I think we're good to go.

PR #282, opened 2016-03-20
Create file 'THEMES'

wiseguys963:
I don't know if there is a list somewhere else, but here it is.
Someone please alphabetize the list before merging to master.
EDIT: Wait, there is. Should have checked docs.


bobthecow:
There's one here: https://github.com/oh-my-fish/oh-my-fish/blob/master/docs/Themes.md

PR #285, opened 2016-03-25
omf update: Report up-to-date packages as such

oranja:
The purpose of this proposal is to generate a more accurate output on omf update. Namely, packages which have no upstream changes now report that they are already up to date.
Since this is a proposal we haven't really discussed before, this is broken into 3 commits (instead of squashed), so it's easier to track the changes and their aims.

Regarding performance, I must admit that I did not time the new code or the old.
Overall the update process should not improve and not deteriorate in terms of performance. The previous git pull is broken apart to a fetch and a merge. Packages without changes skip the merge (and possibly stashing) in return for an additional check. Packages with changes pay just one more check. If git does any more optimizations in combining the fetch and merge when running pull, I am unaware of such and I would appreciate any tips.

### DIFF
--- a/pkg/omf/functions/packages/omf.packages.update.fish
+++ b/pkg/omf/functions/packages/omf.packages.update.fish
@@ -1,19 +1,24 @@
 function omf.packages.update -a name
-  if set target_path (omf.packages.path $name)
-    # Skip packages outside version control
-    not test -e $target_path/.git;
-      and return 0
-
-    if omf.repo.pull $target_path
-      omf.bundle.install $target_path/bundle
-      echo (omf::em)"$name successfully updated."(omf::off)
-      return 0
-    else
-      echo (omf::err)"Could not update $name."(omf::off) 1^&2
-    end
-  else
-    echo (omf::err)"Could not find $name."(omf::off) 1^&2
+  if not set target_path (omf.packages.path $name)
+    echo (omf::err)"Could not find $name."(omf::off) 1>&2
+    return 1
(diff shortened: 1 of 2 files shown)


pkg/omf/functions/packages/omf.packages.update.fish
```diff
@@ -4,15 +4,20 @@ function omf.packages.update -a name
...
-    else
-      echo (omf::err)"Could not update $name."(omf::off) 1^&2
+    omf.repo.pull $target_path
+    switch $status
+      case 0
+        omf.bundle.install $target_path/bundle
+        echo (omf::em)"$name successfully updated."(omf::off)
+        return 0
+      case 1
+        echo (omf::err)"Could not update $name."(omf::off) 1>&2
```
bobthecow: let's just `return 1` here rather than letting it fall back to the one at the bottom. it'll make it more clear.

bobthecow:
I'm :+1: to merge this.

coderstephen:
:+1:

pkg/omf/functions/packages/omf.packages.update.fish
oranja: better now?

derekstavis:
LGTM :+1:

pkg/omf/functions/packages/omf.packages.update.fish
bobthecow: Yep!

PR #260, opened 2016-03-04
init: rewrite init process

derekstavis:
Now use pure globbing to generate 100% valid function and
completion paths, effectively splitting the init process in two
steps, one which paths are added, and other when initialization
is done (sourcing init).

This initialization code introduces a new interface for
`init.fish` hook, which deprecates the previously used event
model. The new interface injects three variables into `init.fish`:
path, package and bundle. This variables can be used by the
package to autoload paths, use bundled files, etc.

Also supports key bindings by sourcing
$OMF_CONFIG/key_bindings.fish and also key_bindings.fish in
packages (plugins and themes) root directories. This is done
when fish_user_key_bindings is called.

PR #286, opened 2016-03-26
Add hooks system for packages

coderstephen:
This pull request adds a new (or rather, revamped) "hooks" system for packages to tap into during the package's lifetime. Similar to Git hooks, OMF hooks are Fish shell scripts located in a `hooks/` directory inside a package. When a particular hook is triggered, the corresponding script is run, if it exists.

The following hooks are implemented:

- `install`; Triggered when the package is first installed via `omf install`. Useful for installing custom dependencies.
- `update`: Triggered when the package is being updated via `omf update`. Useful for updating submodules, etc.
- `uninstall`: Triggered when the package is removed via `omf remove`. Subscribers can use the event to clean up custom resources, etc.

For consistency, the `uninstall` hook is now in the `hooks/` directory, same as the other hooks.

This PR has a weird relationship with #260; it may or may not need to be merged after #260 is. It is based off of `master` though.

Only the English docs are updated; the docs team can update translations separately if this is merged.

### DIFF
--- a/docs/en-US/Packages.md
+++ b/docs/en-US/Packages.md
@@ -53,41 +53,72 @@ Bear in mind that fish lacks a private scope, so if you need to split your packa
 
 # Hooks
 
-Oh My Fish package API offer some lifecycle hooks, which are completely optional. When using `omf new` command some of them are created with sample code.
+Oh My Fish provides a "hooks" system that allows you to write scripts for your package that run when other interesting events occur. Packages can use these hooks to provide advanced installation, custom resource management, etc. Hooks are ordinary Fish scripts named after the event they are triggered by. Most hooks reside in a `hooks` directory inside a package's project directory.
 
 >Hooks that are called at startup time (`init.fish` and `key_bindings.fish`) can slow down shell startup. Be sure to avoid slow code at startup time! Also, if your package doesn't need a hook file, be sure to remove it.
 
-## Initialization
+The working directory inside a hook is always set to the root directory of the package. The hooks Oh My Fish currently supports are listed below:
 
-If you want code to be executed when the package loads, you can add code to `init.fish` file at package's root directory:
+## `init`
+
+The `init` hook is run once when the shell first loads. Scripts to handle this hook should be located at `init.fish` at package's root directory.
+
+Inside this hook, you can access three package-related variables:
+
+* `$package`: Package name
+* `$path`: Package installation path
+* `$dependencies`: Package dependencies
+
+For example, with an `init.fish` script containing the following code:
 
 ```fish
 echo "hello_world initialized"
 ```
 
-Inside this hook runs you can access three package-related variables:
-
-* `$package`: Package name
-* `$path`: Package installation path
-* `$dependencies` : Package dependencies
+you will see the line `hello_world initialized` at the top of the terminal when it is first opened.
 
 Use this hook to modify the environment, load resources, autoload functions, etc. If your package does not export any function, you can still use this event to add functionality to your package, or dynamically create functions.
 
-## Key Bindings
+## `key_bindings`
 
-If your package or theme need to use key bindings, be sure to set them up in `key_bindings.fish`.
+If your package or theme need to use key bindings, be sure to set them up in the `key_bindings` hook. Key binding scripts must be located at `key_bindings.fish` at package's root directory. In this hook you can freely use the [`bind`][fish-bind] command to define custom key bindings.
 
 >Themes can define key bindings too! Oh My Fish will reload key bindings when you switch themes.
 
-## Uninstall
+## `install`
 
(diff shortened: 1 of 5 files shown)


derekstavis:
I always wanted install and update hooks! This hooks are very useful for Git submodule usage or other custom code that runs only once after install/update. I see lots of use cases for this!

I have a fast review of the PR, and I have concerns about the concepts:

* Do we want to move hooks to a entire new directory instead of package root directory, as we use today?
* Calling hooks in a subshell loads entire Oh My Fish again in the subprocess. I don't know whether this can cause issues on update or uninstall hooks.
* I see that the update hook is called twice, so in this case it should be more appropriate to use separate `before_update` and a `after_update` hooks, rather than the same.

Aside from this concerns I love your initiative! Thanks for such PR  🎉

coderstephen:
Thanks for the feedback. Let me address each of your points:

1. Personal preference, but I don't like a messy repo root directory :wink:. My concern actually was to _benefit_ (or at least not _break_) foreign packages. A lot of them still put functions to be loaded on init in the package root. And if they had a function named the same thing as a hook... it would be run as a hook instead of being a function. Only issue then is that `<pkg>/uninstall.fish` still needs to be addressed for BC sake.
2. I was playing around with my implementation and put

   ```fish
   cd $argv[1]
   git submodule update
   ```

   as a test. But this changed *my* current directory. Of course, right? But I didn't think about that. Maybe there's a better way, but to prevent "fiddling with things", I used a subshell. Not really ideal.
3. That's a Git mistake on my part. :scream: It should be triggered only once.

oranja:
I was going to reinforce @derekstavis concerns regarding the double update and the subshell invocation, but you got there first.
Having the hooks in a sub-directory seems like a better idea to me too. It should help to preserve compatibility with other frameworks.

What concerns me the most right now is that I'm still not convinced about use cases.
You both seem to have git submodules in mind, but if submodules are the main use of these hooks and it becomes something we wish to encourage and support, why not just target submodules directly? Like we intend to do with requirements and key-bindings?
Does anyone have an argument for why each package should manage its own submodules (if it becomes something of a habit), or has any other practical use case for these hooks to present?

derekstavis:
I don't believe submodules should be a common thing to deserve a implementation inside the framework.

For example, nvm and rvm plugins should depend on user-installed dependencies, instead of bundling them using submodules.

oranja:
Then the question remains: which type of packages do need submodules? or the other hooks for any purpose?
Until I'm convinced I remain neutral on this. On the one hand change is good. On the other hand growing the code base is not always the best.

coderstephen:
Mostly packages that have third-party shell scripts, like spark or fasd. Using submodules has the advantage of reducing the number of install / update steps, since otherwise you need to download a shell script by hand and put it in the path somewhere. And since it's just shell scripts, compilation, compilers, interpreters, etc. are not a concern.

Now things that need binaries should probably avoid submodules, since it would be better to install the binaries with the system package manager.

coderstephen:
Just pushed a commit that improves the following:

- `source` is used to run hook scripts.
- Instead of `$argv`, the `$package` and `$path` variables are provided for hook scripts.
- The working directory inside a hook script is the package root directory.
- As long as the hook script isn't too naughty, the current user working directory is preserved.

oranja:
The code looks great now IMO.

pkg/omf/functions/packages/omf.packages.hook.fish
```diff
@@ -0,0 +1,20 @@
+function omf.packages.hook -a path hook
```
derekstavis: Not being nitpicky, but `omf.packages.hook.run` sounds more descriptive regarding to the action which the function performs.

pkg/omf/functions/packages/omf.packages.hook.fish
```diff
@@ -0,0 +1,20 @@
...
+  set -l hook_script "$path/hooks/$hook.fish"
+  set package (basename $path)
+
+  if test -e "$hook_script"
+    echo "Running '$hook' hook in package $package..."
```
derekstavis: Wouldn't be better this logging to be under a development flag?

pkg/omf/functions/packages/omf.packages.hook.fish
coderstephen: Fair enough. How about `omf.packages.run_hook`?

derekstavis: Was my second option 😝

coderstephen:
Renamed `omf.packages.hook` function to `omf.packages.run_hook`, rebased, squashed, and removed verbosity from `run_hook`.

derekstavis:
Im 👌 for merging

PR #287, opened 2016-03-28
bugfix: Core update reported failure when up-to-date

oranja:
Due to an oversight in my PR #285, OMF core update reports failure whenever the core is already at the latest revision. This is a simple bugfix for immediate resolution.

As further improvement it's possible to differentiate "updated just now" from "was already up to date" message, but it's not urgent by any means.


bobthecow:
:+1: :shipit:

derekstavis:
Ship it :shipit:

Issue #10, opened 2015-09-04
Start tagging/versioning updates


derekstavis:
We need a way to versionate repository changes via git tags and ensure that `omf update` only fetch and upgrade using tags too. This way we can continue developing on master branch without breaking user installations. It will also make easier to report bugs and track them too.

oranja:
If I may, I like the idea of tagging releases. :)
Is there any way I can help?

p.s.
As an intermediate stage, it's also possible to do all development and testing on a _dev_ branch and only push to _master_ when it's stable.

bpinto:
Hey, that would be great!

We were thinking about having a stable release before we start using tagged releases but I'd like to hear what you have to say about this. Maybe we are wrong :)

oranja:
Naturally any version tag should point to a stable release.
On the upside, I wouldn't say that OMF is not stable, judging by the fact that everything seems to work fine right out of the box. I do, however, think that introducing named releases should be a planned move (and I'm not a huge fan of plans, if it means anything).

My humble suggestions are as follows:

1. First things first - separate _dev_ from _master_.
Fish newcomers are probably also OMF newcomers, and vice versa. If things get rough too soon, most of them will take a u-turn and never come back to either of them (see: vim). Fish is all about love on first sight. So perhaps now things are fine, but as long as anyone who uses OMF gets it straight from the _master_ branch, it will become an impediment. Putting end-users and collaborators on the same branch means that end-users will have to put up with breakages and devs will have to tip toe.
Do development in topic branches, bug-fixes in dev, test from dev, push the ripe fruit to master, or something like that (see: git-flow). This is also not mutually exclusive with version tags, so why not do both anyway?

2. If you want to be taken seriously, I guess you should take yourself seriously. Not too seriously, obviously, but seriously enough to set a milestone and a few modest goals. Looking at the current issues and the way things are, I think OMF can and should strive to hit a milestone and have a proud first release tag.
  1. #152 and #196 lead to the same conclusion: Conventions are great, but packages can't rely on conventions alone anymore. It's cumbersome and not flexible enough. Besides, it makes the code more patch-y. Metadata should be a part of every package. Name, type, and dependencies. At some point there might be a need to add a version field. Descriptions can be optional. It shouldn't be too hard to implement and through this process we also get to do some house cleaning and make sure all plugins are up to date.
[...]

PR #293, opened 2016-03-31
Implement update channels to introduce release versioning

sagebind:
This PR introduces support for release versioning for the core framework to address issue #10. The update process is divided up into different _update channels_; currently the ~~`release`~~ `stable` and `dev` channels are implemented.

Updating the core framework is now split depending on what the current update channel is. If the user is on the `dev` update channel, updates will behave exactly the same way as previous versions. If the user is on the ~~`release`~~ `stable` channel (the default), updating will instead fetch and checkout the newest tag from GitHub.

The user can view, or change, their current update channel with a new command `omf channel`.

In addition, the installer also selects the latest stable version when cloning and installing the core framework from GitHub.

~~_Note:_ Not yet addressed is how existing users will be notified of this change after an update.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,12 @@ Please note that the _`[<url>]`_ portion of the command should be the url to you
 
 Make sure to [send us a PR][omf-pulls-link] to update the registry.
 
+#### `omf channel`
+
+Gets or changes the update channel.
+
+Two channels are available by default: the `stable` channel provides stable updates with the latest tagged version of Oh My Fish, and `dev` which provides the latest changes under development. The update channel currently set determines what version `omf update` will upgrade to.
+
 #### `omf doctor`
 
 Use to troubleshoot before [opening an issue][omf-issues-new].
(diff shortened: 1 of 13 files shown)


coderstephen:
Renamed the `release` channel to `stable`.

coderstephen:
Just updated with the following changes:
- The current installed version is now dynamically determined, so `$OMF_VERSION` is removed and `omf.version` is used instead. The version is based off of the most recent tagged version as well as the commit ID.
- Upgrading from an old version of OMF just got better. The user sees the following message after updating for the first time with the new version:

   ```
   >> You have been switched to the stable release channel of Oh My Fish.
   >> To switch back to the development channel, run `omf channel dev`.
   ```
- Updating the framework now displays a "You are now using Oh My Fish version X." message on success.

sheldon:
I've just reviewed this all and it looks good here, nice work. 🍰 

One thing that would break by merging this up right away is that there are no tagged versions yet. Perhaps we can press a v0.something from the existing version to get things going?

PR #295, opened 2016-04-04
Create fishtory

enricobacis:
This package is just an automated way to install the Bash Style History Substitution (!! and !$)


derekstavis:
We have `bang-bang` which does exactly the same job, with the same code :stuck_out_tongue: 
Give a look at https://github.com/oh-my-fish/plugin-bang-bang

enricobacis:
My fault, thanks! ;)

PR #298, opened 2016-04-13
Add plugin to grep error messages

Shadowigor:
Inspired by [plugin-errno](https://github.com/oh-my-fish/plugin-errno), I took a different approach. The main reason for this was, that I couldn't convert an error number to the abbreviated error label, only to the message. Example: If I enter 'strerrno 22' I only get 'Invalid argument' (actually, it outputs 'Is a directory', which is wrong), but I can't find out that this is called EINVAL. To circumvent this, I'm outputting all three variants. In addition to this, you can also grep for the error messages directly. Have a look at the source [here](https://github.com/Shadowigor/plugin-errno-grep) for examples.

### DIFF
--- a/db/pkg/errno-grep
+++ b/db/pkg/errno-grep
@@ -1,1 +1,1 @@
-.../errno-grep.git
+https://github.com/Shadowigor/plugin-errno-grep.git
--- a/db/pkg/errno-grep
+++ b/db/pkg/errno-grep
@@ -1,0 +1,1 @@
+.../errno-grep.git


oranja:
As a user, I'd rather see just one errno plugin in OMF's database.
@derekstavis, @Shadowigor: maybe you two can merge efforts here?

That's just my opinion, of course.

Shadowigor:
I agree, but I'd suggest that my function would just be an additional function of plugin-errno. The existing functions are more suited for scripting while mine might be better for lookup.

oranja:
Usually when the difference in behavior is about interactive vs. non-interactive (i.e. script), it is easily and conveniently solved with command-line flags. Alternatively, the plugin can identify in which environment it's running. I'm in favor of the former (power to the people).
You brought a great example to the type of stuff that can go wrong when having multiple tools doing the same job (errno 22). When it's just flags controlling the output format the user can be more certain that the main logic stays the same, regardless of output format.

Shadowigor:
You are right. I tried to implement your suggestion [here](https://github.com/Shadowigor/plugin-errno-grep/tree/errno-integration). What do you think?

oranja:
I left you some comments on one of the original commits, have you see them?
I will put the rest of my comments on the new commit.

Shadowigor:
Oh, no, I overlooked them. I restructured my code now and especially changed how the input is handled. You can see it [here](https://github.com/Shadowigor/plugin-errno-grep/commit/fa368d2551da616cef33707aa9a536c75a51c8d6).

derekstavis:
You should definitely give a look at https://github.com/oh-my-fish/plugin-errno :)

derekstavis:
@Shadowigor I've designed `errno` to leverage fish's own autocomplete pager, which have builtin interactive search using <kbd>TAB</kbd> <kbd>TAB</kbd> <kbd>↓</kbd>:

screen shot 2016-07-23 at 11 52 05 pm

And when you start typing:

screen shot 2016-07-23 at 11 52 28 pm

I have tested your package and, as you said, it serves a different purpose and output format, but I'd have no issues in merging your needs and getting `errno` into something that works for both use cases, using some command line switches. What do you think?

Shadowigor:
That's true, but tab-completion works with my version just the same. My suggestion to combine them would be, that if you have an argument like 'ENOMEM', that it has your behavior. You can still change that with a flag, like -f for full or -g for grep, something like that. And if you have an argument that doesn't match an error label, it automatically outputs the grepped output. That would be more or less my tool which would assume the "-n" flag is set if the input is a label. What do you think?

(PS: Make sure you check out the "errno-integration" branch. I just realized, that most of the functionality wasn't merged into master yet.)

derekstavis:
Closing this as now package database is not located at [oh-my-fish/packages-main](https://github.com/oh-my-fish/packages-main)

PR #305, opened 2016-04-19
Fix #303

### LINKED ISSUE #303: omf install removes symlinked bundle
I use a dotfile manager that symlinks `$OMF_CONFIG/bundle`. When I run `omf install <plugin-name>` the link gets replaced by a file. Many dotfile managers use symlinks. It would be useful to change this behavior.

andrewrynhard:
This PR adds the ability to set `bundle` appropriately in the case that it is a symlink.

### DIFF
--- a/pkg/omf/functions/bundle/omf.bundle.add.fish
+++ b/pkg/omf/functions/bundle/omf.bundle.add.fish
@@ -1,8 +1,8 @@
 function omf.bundle.add -a type name_or_url
+  set -l bundle $OMF_CONFIG/bundle
+
   if test -L $OMF_CONFIG/bundle
     set bundle (readlink $OMF_CONFIG/bundle)
-  else
-    set bundle $OMF_CONFIG/bundle
   end
 
   set -l record "$type $name_or_url"
--- a/pkg/omf/functions/bundle/omf.bundle.add.fish
+++ b/pkg/omf/functions/bundle/omf.bundle.add.fish
@@ -1,5 +1,10 @@
 function omf.bundle.add -a type name_or_url
-  set -l bundle $OMF_CONFIG/bundle
+  if test -h $OMF_CONFIG/bundle
+    set bundle (readlink $OMF_CONFIG/bundle)
+  else
+    set bundle $OMF_CONFIG/bundle
+  end
+
   set -l record "$type $name_or_url"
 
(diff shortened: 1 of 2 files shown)


pkg/omf/functions/bundle/omf.bundle.add.fish
```diff
@@ -1,5 +1,10 @@
 function omf.bundle.add -a type name_or_url
-  set -l bundle $OMF_CONFIG/bundle
+  if test -h $OMF_CONFIG/bundle
```
oranja: I'm looking both at the fish website and the man page, and `-h` is undocumented. Maybe `-L` would make it easier for people to understand what it's doing?

pkg/omf/functions/bundle/omf.bundle.add.fish
```diff
@@ -1,5 +1,10 @@
 function omf.bundle.add -a type name_or_url
-  set -l bundle $OMF_CONFIG/bundle
```
oranja: By removing the `-l` flag, you allow `$bundle` to clutter the user's environment.
I suggest keeping this line as default, and then you can change it inside the first `if` case, only if it's a symlink. The second case is then irrelevant.

oranja:
I added some comments on the code, but overall this is even better than what I had in mind.
Thank you!

pkg/omf/functions/bundle/omf.bundle.add.fish
andrewrynhard: `-l` was not used in `omf.bundle.remove.fish`. Would you like the `-l` flag there as well?

oranja: Yes, please.
It's always good to have a new pair of eyes looking at the code. :)

andrewrynhard:
Let me know if I can change anything else.

oranja:
:+1: Looking good.
We'll wait to see what the others think. They are the ones with the commit permissions anyway... ;)

derekstavis:
Cool! I like this fix, very clean and simple. Thanks for implementing!!

coderstephen:
Elegant solution. :+1:

derekstavis:
I think we can merge it :+1:

bobthecow:
:+1:

PR #313, opened 2016-05-03
Streamlined Errors Message

lfiolhais:
When an update fails instead of getting this garble of output:
```
fatal: unable to access 'https://github.com/oh-my-fish/oh-my-fish.git/': Could not resolve host: github.com
Oh My Fish failed to update.
Please open a new issue here → github.com/oh-my-fish/oh-my-fish/issues
fatal: unable to access 'https://github.com/oh-my-fish/theme-default/': Could not resolve host: github.com
Could not update default.
fatal: unable to access 'https://github.com/oh-my-fish/plugin-errno/': Could not resolve host: github.com
Could not update errno.
fatal: unable to access 'https://github.com/lfiolhais/plugin-homebrew-command-not-found/': Could not resolve host: github.com
Could not update homebrew-command-not-found.
fatal: unable to access 'https://github.com/oh-my-fish/plugin-python/': Could not resolve host: github.com
Could not update python.
fatal: unable to access 'https://github.com/lfiolhais/theme-simple-ass-prompt/': Could not resolve host: github.com
Could not update simple-ass-prompt.
fatal: unable to access 'https://github.com/oh-my-fish/plugin-tab/': Could not resolve host: github.com
Could not update tab.
fatal: unable to access 'https://github.com/oh-my-fish/pkg-z/': Could not resolve host: github.com
Could not update z.
```

You simply get this:
```
Oh My Fish failed to update.
Please open a new issue here → github.com/oh-my-fish/oh-my-fish/issues
Could not update default.
Could not update errno.
Could not update homebrew-command-not-found.
Could not update python.
Could not update simple-ass-prompt.
Could not update tab.
Could not update z.
```

### DIFF
--- a/pkg/omf/functions/packages/omf.packages.install.fish
+++ b/pkg/omf/functions/packages/omf.packages.install.fish
@@ -3,7 +3,7 @@ function __omf.packages.install.success
 end
 
 function __omf.packages.install.error
-  echo (omf::err)"Could not install $argv."(omf::off) 1^&2
+  echo (omf::err)"Could not install $argv:$omf_repo_clone_error"(omf::off) 1^&2
 end
 
 function __omf.packages.install.error.already
@@ -37,6 +37,7 @@ function omf.packages.install -a name_or_url
     end
   else
     echo (omf::dim)"Installing $install_type $name"(omf::off)
+    set -g omf_repo_clone_error
 
     if omf.repo.clone $url $OMF_PATH/$parent_path/$name
       omf.bundle.install $OMF_PATH/$parent_path/$name/bundle
@@ -48,8 +49,10 @@ function omf.packages.install -a name_or_url
       end
     else
       __omf.packages.install.error "$install_type $name"
+      set -e omf_repo_clone_error
       return $OMF_UNKNOWN_ERR
     end
+    set -e omf_repo_clone_error
   end
 
   return 0
--- a/pkg/omf/functions/packages/omf.packages.update.fish
+++ b/pkg/omf/functions/packages/omf.packages.update.fish
@@ -8,17 +8,21 @@ function omf.packages.update -a name
   not test -e $target_path/.git;
     and return 0
 
+  # Create omf_repo_pull_error so that we don't have to pass variables
+  # around
+  set -g omf_repo_pull_error
   omf.repo.pull $target_path
   switch $status
     case 0
       omf.bundle.install $target_path/bundle
       echo (omf::em)"$name successfully updated."(omf::off)
     case 1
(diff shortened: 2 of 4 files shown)


oranja:
The tidiness of the second option is clear to see, but only seldom do I find it reasonable to hide a warning or an error from the user.
There are many reasons that a package might fail to update. This specific error provides a fairly concise and accurate description, if you ask me. As long as the signal-to-noise ratio is high, what's wrong with showing the error details?
I will only cast my vote for such a change if some way to get the error is retained. For example with a debug flag or a log file. Yet even then, I'm still not sure that it doesn't just add more work for the rare cases when something breaks.

Either way, this is just my opinion, and the effort is appreciated regardless. Thank you @lfiolhais.

derekstavis:
Hey! Thanks @lfiolhais!

I like the final result. Less noisy. But I also like @oranja thoughts.
Maybe we could meld both ideas and have an output like this one?:

```
Oh My Fish failed to update.
Please open a new issue here → github.com/oh-my-fish/oh-my-fish/issues
Could not update default: Could not resolve host: github.com
Could not update errno: Could not resolve host: github.com
Could not update homebrew-command-not-found: Could not resolve host: github.com
Could not update python: Could not resolve host: github.com
Could not update simple-ass-prompt: Could not resolve host: github.com
Could not update tab: Could not resolve host: github.com
Could not update z: Could not resolve host: github.com
```

(We could also have different colors for the error messages)

lfiolhais:
Hi @oranja, thank you for the input. I do agree with you that the user should be informed of the error however, for new users this becomes more of a nuisance rather than a helpful and informative warning. That being said I like @derekstavis proposal, it is concise, clean and most importantly informative. A flag could be provided to keep raw errors in.

lfiolhais:
This is the solution I came up with. Could you guys give feedback on it? I'm not particularly fond of global variables but that's the only way I found to have a variable across several functions whilst keeping the same value. :disappointed:

lfiolhais:
Closing, as I don't find this to be a good solution to the problem.

oranja:
@lfiolhais, I'm terribly sorry for keeping you hanging like this.
I don't think that the severity and rarity of this issue justifies a global variable to keep error messages, but if you ever feel like coming back to this (either to convince me of the value or with a different approach), do ping me here or on Slack.
I also feel obliged to declare that I'm just a guy hanging around and trying to help, and I don't call any shots around here. In other words: see my arguments as a service, or just ignore them. I don't mean, and don't have the authority, to be a barrier.

The search functionality (#314) is awesome :+1: :clap:
Hope to keep seeing you around!

PR #318, opened 2016-05-13
Bump minimum Fish version in installer (fix #243)

### LINKED ISSUE #243: Bump minimum fish version
We currently depend on `function -V`, a fish 2.2 feature. Looks like it's time to bump minimum fish version to 2.2. For this we need to:

- [ ] Bump minimum fish version in installer
- [x] Add `omf doctor` check for minimum version (Done in #280)

coderstephen:


coderstephen:
Travis, why you no build?!?

PR #324, opened 2016-05-27
Add an autocompletion for fastlane, which is an iOS and Android Automation for Continuous Delivery.

archmagees:
Website: [https://fastlane.tools/](https://fastlane.tools/)


archmagees:
Actually, the commit file should be in the dir ~/.local/share/omf/pkg/shark/completions , but I don't know where to put it in. So please reset the file's position. And tell me the right way to pull request better. 
Thank you.

archmagees:
Now I put the file in the fish-shell directly. The former place is not a good place to place it. 
Sorry to bother you!

coderstephen:
@archmagees No problem. It looks like you found the proper place to put it. Closing this PR then.

PR #330, opened 2016-06-01
Inspect $FISH_VERSION variable instead of $version

lfiolhais:
Per fish-shell/fish-shell@95635a5

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -54,8 +54,9 @@ function report -a what message
 end
 
 function fish_version_compatible
-  set -l major (echo $version | cut -d. -f1)
-  set -l minor (echo $version | cut -d. -f2)
+  set -q FISH_VERSION; or set -l FISH_VERSION $version
+  set -l major (echo $FISH_VERSION | cut -d. -f1)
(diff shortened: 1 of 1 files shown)


derekstavis:
Hey, thanks for the PR! :+1:

lfiolhais:
No problem, we would have to do this when the next version shipped so we might as well do it now.

bobthecow:
:+1:

coderstephen:
Should the installer check both `$FISH_VERSION` and `$version`? I don't recall `$FISH_VERSION` being defined for Fish < 2.3.0.

lfiolhais:
Didn't think of that @coderstephen. I'll add it now.

bin/install
```diff
@@ -54,8 +54,8 @@ function report -a what message
 end
 
 function fish_version_compatible
```
bobthecow: set -q FISH_VERSION; or set -l FISH_VERSION $version

bobthecow:
:+1: mind rebasing so we can get a clean merge?

scorphus:
:+1: let us go for it

PR #334, opened 2016-06-03
Fix theme key bindings

derekstavis:
This PR fixes the way themes key bindings are treated.

Previous to this PR all installed themes key bindings would be loaded. This PR makes startup code load only current theme key bindings, when existent. 

Also, previously, key bindings wouldn't be reloaded on theme change. This reloads key bindings file when it's found on the theme.

----

@tannhuber, I see your themes use key bindings. Would you like to help us test this PR?

To enable automatic key binding load on themes, move all `bind` calls to `key_bindings.fish` file in theme's repository root directory.

This PR should make `omf t dangerous; omf t budspencer` use `budspencer` key bindings.

### DIFF
--- a/init.fish
+++ b/init.fish
@@ -25,13 +25,15 @@ emit perf:timer:finish "Oh My Fish init installed packages"
 functions -q fish_user_key_bindings
   and functions -c fish_user_key_bindings __original_fish_user_key_bindings
 # Override key bindings, calling original if existent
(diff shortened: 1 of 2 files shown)


tannhuber:
I tested it successfully. All checks have passed. Thank you.

derekstavis:
Thanks for the testing @tannhuber

PR #343, opened 2016-06-08
require: fix completions autoload

derekstavis:
This PR fixes a typo in `require` function that made completions not be autoloaded at shell startup.


bobthecow:
:+1:

PR #346, opened 2016-06-10
Update peco repo

pedrosnk:
When I upgraded my omf peco stopped to work I had to change the repo to this one that seems the one the is currently maintainable.


derekstavis:
Thanks for the fix!

PR #348, opened 2016-06-15
Evaluate OMF_PATH at init time to fix #136

### LINKED ISSUE #136: install script hardcodes home directory
I'm not sure if this is intended or not, but i ended up with

```bash
set -gx OMF_PATH /home/johnny/.local/share/omf
#set -gx OMF_CONFIG /home/johnny/.config/omf
```

when I expected:
```bash
set -gx OMF_PATH $HOME/.local/share/omf
#set -gx OMF_CONFIG $HOME/.config/omf
````

coderstephen:
If we want to evaluate `$OMF_PATH at init time instead of install time, this is how we can do it. A fair reason we might want to do this is it makes putting `config.fish` into a dotfiles repo make more sense. Currently hardcoding the paths makes the config file machine-specific.

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -109,7 +109,7 @@ function install_omf
   report progress "Adding startup code to fish config file..."
 
   set template "templates/config.fish"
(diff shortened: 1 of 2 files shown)


derekstavis:
👍

PR #350, opened 2016-06-15
Fix #189

### LINKED ISSUE #189: Calling oh-my-fish installation from a script never returns control to that script
I have a python script I run when I install Ubuntu (it installs some apps I need etc). And it installs oh-my-fish as well: 
```
system('sudo apt-get --assume-yes install fish')
system('curl -L github.com/oh-my-fish/oh-my-fish/raw/master/bin/install | fish')
... # other installation code
```
I run this script in terminal right after installing Ubuntu (15.10). But the problem is that oh-my-fish installation starts *fish* process and never terminates it, so my script can't continue. I get the following output:
```
Installing Oh My Fish to /home/d/.local/share/omf...
Cloning master from https://github.com/oh-my-fish/oh-my-fish.git...
Existent config.fish found at /home/d/.config/fish
↳ Moving to /home/d/.config/fish/config.1449221085.copy
Adding startup code to fish config file...
Building Oh My Fish configuration...
Installing package gi
✔ package gi successfully installed.
d@d ~> 
```

In fact omf installation starts another *fish* process within itself and never teriminates that, so I even don't see the success report. So the code `report success "Installation successful!"` also wasn't executed.

*htop* outputs the following tree: 
```
bash
↳  python
   ↳  fish 
      ↳ fish
```

coderstephen:
`source` to finish installing instead of running a subshell.

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -126,7 +126,8 @@ function install_omf
   test -f "$OMF_CONFIG/bundle"; or echo "theme default" > "$OMF_CONFIG/bundle"
   test -f "$OMF_CONFIG/theme";  or echo "default" > "$OMF_CONFIG/theme"
 
-  fish -c "omf install"
+  source "$fish_config_file"
+  omf install
 end
 
 function main


derekstavis:
Closing PR as the upcoming #361 will obsolete it.

PR #352, opened 2016-06-25
Add "sudope" to the registry

oranja:
With Fish 2.3 and key binding support from OMF, this plugin can do its business elegantly.


derekstavis:
Join the organization! `https://github.com/oh-my-fish/plugin-sudope`

oranja:
Thanks! Repository copied over and I will now remove the original to avoid confusion.

PR #357, opened 2016-07-07
Add "eden" to theme registry.

amio:
An evolution of theme "edan" 😄 
https://github.com/amio/fish-theme-eden

eden

### DIFF
--- a/db/themes/eden
+++ b/db/themes/eden
@@ -1,0 +1,1 @@
+https://github.com/amio/fish-theme-eden
--- a/db/themes/eden
+++ b/db/themes/eden
@@ -1,1 +1,1 @@
-https://github.com/amio/fish-theme-eden
+https://github.com/oh-my-fish/theme-eden


derekstavis:
Hey @amio, would you like to join Oh My Fish organization by moving your repository to the org? If you'd like, I can allocate a repo for you and make you an admin. Then you can update the PR and I merge it 👍

amio:
Thanks for your reminder, while I prefer to keep it still :P 

I'd like to remain neutral to frameworks. Although there was some argument between developers, but for me, I appreciate all your contributions to the community, so I want to support all these works (not just one) as much as I can. Besides, I thought it may not be very suitable to have a installing guide with fisher, in an omf project ╮(╯_╰)╭

Thanks again 😁

derekstavis:
giphy

No problem @amio!!

amio:
👍 for the great idea: "aggregate Fish related stuff into a single place people can search." and the registry approach. Furthermore, the registry maybe deserves a standalone repo for that, what do you think?

KhaledMohamedP:
@amio I do support adding a new repo for the registry

amio:
@derekstavis Would you helping allocate a repo for this?

sagebind:
@amio Oops, we should have merged this I think...

> Would you helping allocate a repo for this?

Did you want to keep the theme under your account? That was fine -- we didn't mean to reject it.

amio:
Thank you @sagebind ! And if that's fine too, I prefer to keep the theme repo under my account by now :P

sagebind:
@amio The only problem I forsee is that themes don't always install correctly if the repo is not prefixed with `theme-` -- this is currently a limitation in how OMF detects themes vs plugins.

You will need to rename your repo first before submitting, I think. :/

amio:
Hmm, I understand that 🤔

What about making a fork from `amio/fish-theme-eden` to `oh-my-fish/theme-eden`? I'll keep those two repo in sync; or just allocate a new repo `oh-my-fish/theme-eden` and I'll do the rest.

Both fine by me :D

sagebind:
@amio OK, I'll make a fork, since that seems reasonable to me. I can give you permissions on the fork so that you can update it whenever you like.

You'll need to change this commit to use the URL `https://github.com/oh-my-fish/theme-eden`, then we can merge.

amio:
Thank you @sagebind 😃 I'll update this soon.

amio:
It seems I cannot rename the repo, could you help me with this? @sagebind

sagebind:
@amio Oops, sure.

amio:
@sagebind Both repo & PR updated. Waiting for the confirmation from CI :)

sagebind:
@amio Travis is slow today...

Issue #281, opened 2016-03-20
install overwrites existing ~/.config/fish/config.fish


diggit:
This is really annoying bug. I had already tweaked `~/.config/fish/config.fish` and installation overwrote it. Not a single question and everything is gone :/

coderstephen:
If everything went properly, the installer should have backed up `~/.config/fish/config.fish` to `~/.config/fish/config.fish.TIMESTAMP.copy`. Double-check if there is a backup put into the config folder. If not, the installer may not have run properly, or maybe has a bug.

diggit:
Oh thanks, there is backup file `config.1458508741.copy`. You saved me!

bobthecow:
The installer _should_ have let you know it was moving that to a backup… I'm curious what might have made that more obvious? Did it just get lost in the pile of things the installer says?

https://github.com/oh-my-fish/oh-my-fish/blob/master/bin/install#L74-L75

diggit:
I had problems with fonts. It made terminal uneasy to read. I am probably too used to new files being renamed to .pcanew and not touching changed ones, from Arch. My fault, sorry for false alarm.

JerrySievert:
i also had this issue, which made me uninstall OMF completely.

it should at least warn that it is overwriting the configuration that we worked hard to get, not just make a copy of it.

this is bad behavior.

coderstephen:
The installer is supposed to print a message indicating if it is going to move the existing config, and where it is putting it. Did the message not work for you?

@JerrySievert What behavior would you prefer? We're open to suggestions on how to improve.

JerrySievert:
@coderstephen i think the best would be to append to the config, not completely overwrite it.

the analogy i can up with during a discussion on irc:

"Welcome back to your room, we painted the walls exactly as you wanted, and we think it looks great!  Oh, we went ahead and rented a storage unit and took all of your furniture to it for you.  We decided that the room would look better if it were empty.  I guess you can move the furniture back in, but we're not going to help you with that."

i did not see the message, it could have been there?  it definitely wasn't obvious.

i uninstalled and moved to `fisher install bobthefish`, and it was a much better and cleaner experience.

coderstephen:
> i did not see the message, it could have been there? it definitely wasn't obvious.

[...]

Issue #274, opened 2016-03-15
Installing omf overwrites theme to default


kaelumania:
When I install omf and already have an `~/.config/omf/theme` file which was linked from my `.dotfiles` folder. The the installation overwrites the theme file and sets the theme to default.

derekstavis:
Ah, nice catch! Thanks for reporting :+1:

Issue #267, opened 2016-03-09
Fixing file behaviors


michaelperrygoodman:
Hi. Apologies for how this comes off as I just lost hours of work because of the 'omf destroy' command removing files it had no business touching. Here's the setup (OSX):

I want to contain my customized fish stuff in my Google Drive folder, as well as having it exist as a submodule to my dotfiles repo. I am also trying to run OMF. So I (perhaps foolishly) pointed my 'custom OMF configuration path' to the standard '$HOME/.config/fish' directory. This dir is actually a symlink to my GDrive folder with the same name and structure. It contains a functions/ directory with aliases.fish, fish_prompt.fish and a few other files I source (because I guess I don't really understand how this folder is supposed to be autoloaded by OMF).

Some of the files were not loading correctly and I could not understand why, so I decided to destroy and reinstall OMF. I realize it makes backups when you do an initial install, which I already was not fond of since I do not think it should be touching anything besides the config.fish. And again foolishly removing these backups (specifically for config.fish and functions/fish_prompt.fish). Finally, when I performed 'omf destroy' it did precisely what that command implies and deleted both of these files. It also, however, somehow managed to revert the functions/aliases.fish file to its original state, which destroyed quite a lot of work I had been doing.

So, I guess what I'm wondering here is if there isn't some way to make this "custom" settings folder independent entirely of OMF so it doesn't wreck my work. Perhaps some safeguard checking when working with custom folders? And instead of flat out deleting files, just leaving them alone? I want my custom configurations and functions to work along side OMF, not inside of it, or being managed by it in any way. Am I missing some obvious part of how fish and OMF work? I know I blundered this quite a bit by not taking steps to make sure my data wouldn't be destroyed, but I didn't quite expect this. Thanks for any help.

bobthecow:
It _should_ only remove  `$OMF_PATH`, which is the core OMF library, and restore your backed up files to `~/.config/fish/config.fish` and `~/.config/fish/functions/fish_prompt.fish`.

The latter is destructive, though, and it should probably make a fresh backup before replacing them with your older original files, in case you've customized them after installing OMF.

oranja:
Hi Michael,

First let's address your data: I am pretty certain that GDrive both keeps file versions and has a trash can that you can restore your files from before they're permanently lost. Older versions and deleted files are kept for 30 days, from a quick check.
I recently restored a folder with some 3000 files that was deleted in one moment due to a mistake with a synching software.

[...]

PR #361, opened 2016-07-13
Refactor installer to support offline installs and more control over user data

coderstephen:
Completely refactors the installer. The code is more organized and should be more reliable and consistent.

The new installer supports:
- offline installation (#167, #185)
- interactive questions and acknowledgements
- preserving and ignoring user's config files (#267, #274, #281)
- bootstrap code in `config.fish` is no longer required for Fish 2.3.0+, uses `conf.d` instead
- customizable install paths via command flags

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -1,158 +1,524 @@
 #!/usr/bin/env fish
-#
-# USAGE
-#   #1: curl -L https://github.com/oh-my-fish/oh-my-fish/raw/master/bin/install | fish
-#   #2: curl -L https://github.com/oh-my-fish/oh-my-fish/raw/master/bin/install > install; and fish install
-#   #3: env OMF_CONFIG=~/.omf curl -L https://github.com/oh-my-fish/oh-my-fish/raw/master/bin/install | fish
-#
-# ENV
-#   XDG_DATA_HOME       Base directory (~/.local/share)
-#   XDG_CONFIG_HOME     Base configuration directory (~/.config)
-#
-#   - See XDG Base Directory Specification:
-#   - https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
-#
-#   FISH_CONFIG       Fish shell configuration file
-#
-#   OMF_PATH          Oh My Fish installation directory
-#   OMF_CONFIG        Oh My Fish configuration directory
-#
-#   OMF_REPO_URI      Oh My Fish source git repository
-#   OMF_REPO_BRANCH   Oh My Fish source repository branch
+# Oh My Fish installer. See `install --help` for usage.
 
-
-set -q XDG_DATA_HOME;   or set XDG_DATA_HOME "$HOME/.local/share"
-set -q XDG_CONFIG_HOME; or set XDG_CONFIG_HOME "$HOME/.config"
-
-set -q FISH_CONFIG;     or set FISH_CONFIG "$XDG_CONFIG_HOME/fish"
-
-set -q OMF_PATH;        or set OMF_PATH "$XDG_DATA_HOME/omf"
-set -q OMF_CONFIG;      or set OMF_CONFIG "$XDG_CONFIG_HOME/omf"
-
+# Set environment options.
 set -q OMF_REPO_URI;    or set OMF_REPO_URI "https://github.com/oh-my-fish/oh-my-fish"
 set -q OMF_REPO_BRANCH; or set OMF_REPO_BRANCH "master"
 
-set OMF_FISH_MIN_VER 2 2 0
 
-function available -a name
-  type "$name" >/dev/null 2>&1
-end
+function main
+  # Set default settings
+  set -q XDG_DATA_HOME
+    and set -g OMF_PATH_DEFAULT "$XDG_DATA_HOME/omf"
+    or set -g OMF_PATH_DEFAULT "$HOME/.local/share/omf"
 
-function report -a what message
-  switch $what
-  case 'progress'
-    set_color yellow
-  case 'success'
-    set_color green
-  case 'error'
-    set_color red
-    printf "$message\n"
-    exit 1
+  set -q XDG_CONFIG_HOME
+    and set -g CONFIG_PATH "$XDG_CONFIG_HOME"
+    or set -g CONFIG_PATH "$HOME/.config"
+
+  set -g OMF_CONFIG_DEFAULT "$CONFIG_PATH/omf"
+  set -g FISH_CONFIG "$CONFIG_PATH/fish"
+
+  set -g OMF_PATH "$OMF_PATH_DEFAULT"
+  set -g OMF_CONFIG "$OMF_CONFIG_DEFAULT"
+
+  # Ensure repository URL ends with .git
+  set OMF_REPO_URI (echo $OMF_REPO_URI | command sed 's/\.git//').git
+
+  # Parse command args
+  while set -q argv[1]
+    switch "$argv[1]"
+      case --help -h '/?'
+        echo "Usage: install [options]
+  Install Oh My Fish
+
+Options:
+  --config=<path>         Put config in a specific path (default is $OMF_CONFIG_DEFAULT).
+  --help, -h              Show this help message.
+  --noninteractive        Disable interactive questions (assume no, use with --yes to assume yes).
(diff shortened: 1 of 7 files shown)


bin/install
```diff
@@ -1,158 +1,449 @@
...
+  return 0
 end
 
-function fish_version_compatible
-  set -q FISH_VERSION; or set -l FISH_VERSION $version
-  set -l major (echo $FISH_VERSION | cut -d. -f1)
-  set -l minor (echo $FISH_VERSION | cut -d. -f2)
 
-  return (test $major = $OMF_FISH_MIN_VER[1] -a $minor -ge $OMF_FISH_MIN_VER[2])
+# Add an exit hook to display a message if the installer aborts or errors.
```
derekstavis: Cool idea!

bin/install
```diff
@@ -1,158 +1,449 @@
...
+    set -l path (command dirname (command dirname (builtin status -f)))
+
+    if is_install_dir "$path"
+      set OFFLINE_PATH "$path"
+    # Try using the current working directory as the source.
+    else if is_install_dir "$PWD"
+      set OFFLINE_PATH "$PWD"
+    else
+      # We tried our best.
+      abort "Could not find local installation"
```
derekstavis: Maybe `"Could not find local installation source"` sounds better?

bin/install
```diff
@@ -1,158 +1,449 @@
...
+    type -f -q $cmd
+      or abort "Command '$cmd' not found"
   end
+end
+
+
+# Assert that a minimum required version of Fish is installed.
+function assert_fish_version_compatible -a required_version
+  set -l installed_version (get_fish_version)
+    and is_version_compatible $required_version $installed_version
```
derekstavis: `get_fish_version` should `return 1` for this `and` clause make sense. This change should catch the edge case of neither `FISH_VERSION` and `version` variables exist.

bin/install
```diff
@@ -1,158 +1,449 @@
...
-  end
+# Generates the bootstrap code used to initialize Oh My Fish on shell startup.
+function generate_bootstrap -a config_path
+echo '# Path to Oh My Fish install.
+set -q XDG_DATA_HOME
+  and set -gx OMF_PATH "$XDG_DATA_HOME/omf"
+  or set -gx OMF_PATH "$HOME/.local/share/omf"
+
+# Customize Oh My Fish configuration path.
+#set -gx OMF_CONFIG "'$config_path'"
```
derekstavis: Hm, is this double quoting needed: `"'$config_path'"`?

bin/install
```diff
@@ -1,158 +1,449 @@
...
+
+# Checks if OMF is set up properly and working.
+function is_omf_loaded
+  command fish -c "omf --version" > /dev/null ^&1
+end
+
+
+# Tests if the right-hand side version is equal to or greater than the left-hand side version.
+function is_version_compatible -a lhs rhs
+  test -n "$lhs" -a -n "$rhs" -a "$lhs" = (printf "$lhs\n$rhs\n" | command sort -t '.' -k 1,1 -k 2,2 -k 3,3 -k 4,4 -g)[1]
```
derekstavis: I found this implementation confusing.

Splitting stuff into variables help making it more readable:

```
set -l sorted (printf "$lhs\n$rhs\n" | command sort -t '.' -k 1,1 -k 2,2 -k 3,3 -k 4,4 -g)
test -n "$lhs" -a -n "$rhs" -a "$lhs" = $sorted[1]
```

Still I find the `sort` command a little obscure and prefer the old index comparison method.

bin/install
coderstephen: It isn't double-quoting; it's printing a double quote, exiting the string, printing `$config_path`, opening the string again, and printing another double quote.

derekstavis: Oh sure. Kinda tricky! :stuck_out_tongue:

derekstavis:
What about using `http://get.oh-my.fish` URL?

derekstavis:
Still missing is the uninstall method compatible with fish 2.3.
Making `omf destroy` call `bin/install --uninstall` would do the trick :)

derekstavis:
After installing, running the `fish install --help` is not possible?

screen shot 2016-07-23 at 9 08 12 pm

derekstavis:
When installing to a path using `fish install --offline=/tmp` and the source isn't available, I get some weird messages 👀 

screen shot 2016-07-23 at 9 11 54 pm

derekstavis:
When installing to a path using `fish install --offline=/tmp` and the source isn't found, I get some weird messages :eyes:

coderstephen:
@derekstavis Good suggestions. I just pushed an update that includes/fixes all of them! :+1:

pkg/omf/functions/core/omf.destroy.fish
```diff
@@ -1,38 +1,15 @@
...
-
-  if test "$OMF_PATH" != "$HOME"
-    rm -rf "$OMF_PATH"
-  end
+  # Run the uninstaller
+  fish "$OMF_PATH/bin/install" --uninstall "--path=$OMF_PATH" "--config=$OMF_CONFIG"
 
-  set -q CI; or exec fish < /dev/tty
+  # Start a new OMF-free shell
+  exec fish < /dev/tty
```
derekstavis: Shouldn't this cause an issue with Travis?

bin/install
```diff
@@ -1,158 +1,502 @@
...
+source \$OMF_PATH/init.fish"
+end
+
+
+# Uninstalls an existing OMF installation.
+function uninstall_omf
+  is_install_dir "$OMF_PATH"
+    or abort "No installation detected at $OMF_PATH"
+
+  say "Warning: This will uninstall Oh My Fish and all plugins and themes installed in $OMF_PATH. Your configuration will not be modified. You will need to remove any Oh My Fish startup code from $FISH_CONFIG/config.fish."
```
derekstavis: I think we could add some space between the messages:

```
Warning: This will uninstall Oh My Fish and all plugins and
themes installed in /Users/derek/.local/share/omf.

Your configuration will not be modified. You will need to remove any
Oh My Fish startup code from /Users/derek/.config/fish/config.fish.
```

And also use bold in the last paragraph

bin/install
```diff
@@ -1,158 +1,502 @@
...
+
+  confirm_yes "Are you sure you want to continue?"
+  say "Uninstalling from $OMF_PATH..."
+
+  # Remove the core framework
+  command rm -rf "$OMF_PATH/.git"
+    and command rm -r "$OMF_PATH"
```
derekstavis: We should use `rm -rf` here to avoid git directory permission confirmations:

```
⋊> /p/tmp fish install                                                                                                                                14:02:08
Existing installation detected at /Users/derek/.local/share/omf
Would you like to remove the existing installation? (y/N): y
Warning: This will uninstall Oh My Fish and all plugins and themes installed in 
/Users/derek/.local/share/omf. Your configuration will not be modified. You 
will need to remove any Oh My Fish startup code from 
/Users/derek/.config/fish/config.fish.
Are you sure you want to continue? (y/N): y
Uninstalling from /Users/derek/.local/share/omf...
override r--r--r--  derek/staff for /Users/derek/.local/share/omf/pkg/apt/.git/objects/07/70011cff189b773de7eeedce44cb60ab6c2931?
```

bin/install
```diff
@@ -1,158 +1,502 @@
...
 
-function fish_version_compatible
-  set -q FISH_VERSION; or set -l FISH_VERSION $version
-  set -l major (echo $FISH_VERSION | cut -d. -f1)
-  set -l minor (echo $FISH_VERSION | cut -d. -f2)
 
-  return (test $major = $OMF_FISH_MIN_VER[1] -a $minor -ge $OMF_FISH_MIN_VER[2])
+# Add an exit hook to display a message if the installer aborts or errors.
+function on_exit -p %self
+  if test $status -ne 0 -a $status -ne 2
```
derekstavis: Hm.. I'm not sure sure this works. What I know is that you can

```
echo $argv | read -l event group exit_status
```

To get more information.

coderstephen: `$status` is indeed usable here since the process I'm handling the exit of is myself, so `$status` is still the most recent exit status. `$argv` is probably more correct though. I shall change it.

PR #370, opened 2016-07-24
docs: improve headers, use flag emojis for translations

derekstavis:


bobthecow:
This should use title text to be accessible :)

Issue #344, opened 2016-06-09
Add git shortcuts


aayushkapoor206:
I recently moved from zsh to fish and the thing I miss the most are the git shortcuts...
Such as
gst for git status
gup for git pull --rebase
gp for git push
etc...

I couldn't find any plugin that did this... Can it be enabled by default?

lfiolhais:
Generally, plugins that only setup aliases are frowned upon as they are too personal. As far as I'm aware there are no plugins of that type in the database. Yet, you can create your own aliases in `init.fish` with the `alias` builtin. Like so, `alias gst="git status"`.

nwykes:
I have the following in git_alias.fish which I source from init.fish:

```
alias gd="git diff"
alias gdc="git diff --cached"
alias ga="git add"
alias gca="git commit -a -m"
alias gcm="git commit -m"
alias gbd="git branch -D"
alias gst="git status -sb --ignore-submodules"
alias gm="git merge --no-ff"
alias gpt="git push --tags"
alias gp="git push"
alias grs="git reset --soft"
alias grh="git reset --hard"
alias gb="git branch"
alias gcob="git checkout -b"
alias gco="git checkout"
alias gba="git branch -a"
alias gcp="git cherry-pick"
alias gl="git lg"
alias gpom="git pull origin master"
```

oranja:
Even in `zsh` I am pretty certain that the git aliases are not enabled by default, and  @lfiolhais is right. Aliases are personal and so if we accept one git aliases plugin, it will be joined immediately with 10 other slightly or completely different variants.
You are left with two options:

1. As stated above, you can have your aliases set in `init.fish` (or some file sourced from it).

2. If you insist, create a plugin and share it with friends and colleagues. The plugin will not enter the official registry, but it's still possible to install and manage (update/remove) a plugin with the full url of the repository.
e.g.: `omf install https://github.com/aayushkapoor206/omf-plugin-gitaliases.git`
and `omf remove gitaliases`

derekstavis:
Hi @aayushkapoor206, you can have a look at my own minimal aliases:

https://github.com/derekstavis/plugin-minimal-git-aliases

Give a look at the names I'm using:

https://github.com/derekstavis/plugin-minimal-git-aliases/blob/master/init.fish

KhaledMohamedP:
where is the `init.fish` you guys are referring too, can someone please provide me the `path`?

lfiolhais:
[...]

PR #372, opened 2016-07-26
Add plugin git-aliases.

irstacks:


oranja:
Hey @irstacks ,
Thank you for your suggestion, but there are two problems with this contribution:

1. As a policy, we try to avoid adding collections of aliases as plugins to the official OMF plugin database. There is nothing wrong with these packages, and they often prove helpful, but they are also fairly personal. Since we don't want to accept every variant of the same alias collection and we also don't want to decide for everyone else which collection is "best", we simply avoid all of them.
If you publish a github repository of your own and make it work with OMF, people can use it even if it's not in OMF's database. Please see #344 for a conversation in this matter.

2. Technically, this is not how you submit a package to OMF's database. First you need to create a dedicated repository for your plugin. Then you can submit a Pull Request that only adds the URL of your repository to a simple text file under `db/pkg`. Detailed instructions are available here: [docs/en-US/Packages.md](https://github.com/oh-my-fish/oh-my-fish/blob/master/docs/en-US/Packages.md)
Note that:
  [a] we can allocate a repository under the Oh-My-Fish organization.
  [b] the repository should have a `omf-plugin-` or `plugin-` prefix to be installed with OMF)

irstacks:
Got it! Thanks @oranja.

PR #384, opened 2016-08-12
Add GPG Package for OMF

benkutil:
Instead of manually adding this to your fish config, I’ve wrapped [kvs’](https://github.com/kvs) [gnupg script](https://github.com/kvs/dotfiles/blob/master/fish/gnupg.fish) as an omf plugin.

### DIFF
--- a/db/pkg/gpg
+++ b/db/pkg/gpg
@@ -1,0 +1,1 @@
+https://github.com/benkutil/omf-pkg-gpg


oranja:
**omf-pkg-gpg/functions/omf-gpg.fish** is not needed. I'd remove it.
A more appropriate use for the **functions/** folder is to to move `__refresh_gpg_agent_info` into a **functions/__refresh_gpg_agent_info.fish** file.
I'd also set move the **~/.gnupg/.gpg-agent-info** path into a "constant", or even move the entire `gpg-agent --daemon --write-env-file ~/.gnupg/.gpg-agent-info >/dev/null` line into a tiny function.

Of course, I might have taken it too far. Take what makes sense and leave the rest.

derekstavis:
Closing this as now package database is located at [oh-my-fish/packages-main](https://github.com/oh-my-fish/packages-main)

PR #387, opened 2016-08-14
Basic infra for mercurial support

zahanm:
I'm using these utils in the oh-my-fish/theme-default#7 to show a basic branch / bookmark prompt for hg repos for the default OMF install.

I didn't use any `hg` commands because the overhead of starting up python is too much for an operation on every shell.
ref: https://www.mercurial-scm.org/pipermail/mercurial/2014-November/047694.html
ref2: http://patrickoscity.de/blog/building-a-fast-mercurial-prompt

Not sure if this belongs in a plugin, but it seems up there with basic git functionality to me. (Yes, I use mercurial quite a bit.)

Not doing anything else fancy like dirty working dir etc out of laziness, though I assume looking at the files in `$repo/.hg/` I can do that too.

re: test plan, I'm using this locally quite happily. So I guess it works. :)

### DIFF
--- a/init.fish
+++ b/init.fish
@@ -10,7 +10,7 @@ emit perf:timer:start "Oh My Fish initialisation"
 # Read current theme
 read -l theme < $OMF_CONFIG/theme
 # Prepare Oh My Fish paths
-set -l core_function_path $OMF_PATH/lib{,/git}
+set -l core_function_path $OMF_PATH/lib{,/git,/hg}
 set -l theme_function_path {$OMF_CONFIG,$OMF_PATH}/themes*/$theme{,/functions}
 # Autoload core library
 set fish_function_path $fish_function_path[1] \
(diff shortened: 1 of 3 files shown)


coderstephen:
@zahanm Thanks for the idea, but we're trying to remove VCS-related functions out of the core framework, since it is better suited to be a plugin. In fact, [the official vcs plugin](https://github.com/oh-my-fish/plugin-vcs) supports Mercurial IIRC. The correct solution I believe would be to modify the default theme to use the vcs plugin to do its display instead.

zahanm:
Okay, sounds like this isn't quite what you want.

derekstavis:
@zahanm if you'd like to contribute your work into VCS plugin you are more than welcome! You can checkout a [better version of default theme from my own fork](https://github.com/derekstavis/theme-default) -- it's already integrated into VCS plugin. As I'm not a Mercurial user, my implementation was very simple. We plan to deprecate `lib/git` from Oh My Fish core and move default theme to VCS very soon.

Issue #398, opened 2016-09-05
Command 'basename' not found during installation


mdjarv:
When I try to install omf I get the following error:

```shell
 ~> curl -L http://get.oh-my.fish | fish
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   154  100   154    0     0    427      0 --:--:-- --:--:-- --:--:--   426
100   140  100   140    0     0    160      0 --:--:-- --:--:-- --:--:--   160
100 15336  100 15336    0     0  13889      0  0:00:01  0:00:01 --:--:-- 13889
type: invalid option -- 'q'
Install aborted: Command 'basename' not found
```

I have checked and basename is available under /usr/bin/basename

* Operating System: Ubuntu 14.04.5 Server
* Fish Version: 2.0.0
* Git version 1.9.1
* Basename version 8.21

mdjarv:
Found the culprit, this is where the script failed in the install script (line 449, inside function `assert_cmds`):
```fish
type -f -q $cmd
```

`type -q` is not valid in fish 2.0, and if I remove that parameter I get a more verbose error.

```fish
~> ./install
basename is /usr/bin/basename
cp is /bin/cp
cut is /usr/bin/cut
date is /bin/date
dirname is /usr/bin/dirname
fish is /usr/bin/fish
fold is /usr/bin/fold
git is /usr/bin/git
head is /usr/bin/head
mkdir is /bin/mkdir
mv is /bin/mv
rm is /bin/rm
sed is /bin/sed
sort is /usr/bin/sort
tar is /bin/tar
tr is /usr/bin/tr
Install aborted: Fish version 2.2.0 or greater required; you have 2.0.0
```

PR #399, opened 2016-09-05
assert fish version before asserting commands

mdjarv:
A suggestion for resolving issue #398, this ensures that the correct error message is displayed if the users environment does not meet the requirements

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -72,8 +72,8 @@ Options:
   end
 
   # Ensure the environment meets all of the requirements.
-  assert_cmds
   assert_fish_version_compatible 2.2.0
+  assert_cmds
   assert_git_version_compatible 1.9.5
   assert_interactive
 
--- a/bin/install
+++ b/bin/install
@@ -73,8 +73,8 @@ Options:
 
   # Ensure the environment meets all of the requirements.
   assert_fish_version_compatible 2.2.0
-  assert_cmds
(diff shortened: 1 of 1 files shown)


oranja:
Good job!

 I'd go even further as to remove the checks for `fish` and `git` in `assert_cmds`.
No point in checking these twice. Also if someone gets so far with the installer she probably has `fish` installed, no?

There will still be an environment which will make the installer fail (delete `sort`, for example), but there's only so much we can do, unless we revert to some backwards-compatible check.
Then `assert_cmds` will be called first but use `type` without `-q`  or spawn a `which` for each check.

(These pondering are intended for everyone, not just @mdjarv)

mdjarv:
I agree, to get anywhere in the script the user has to have some version of fish already, and git is asserted with `type -f -q git` in the `assert_git_version_compatible` so those two checks in `assert_cmds` feel redundant and unneccesary.

I'll update the PR with these changes and we'll take the discussion from there.

> ... unless we revert to some backwards-compatible check.
> Then `assert_cmds` will be called first but use `type` without `-q` or spawn a `which` for each check.

The script is currently aborting on the first error it finds during `assert_cmd`. Considering how few dependencies there actually are (fish, git and a few shell-commands that are very common) I don't see any issues with following this pattern when checking fish and git versions as well.

Issue #386, opened 2016-08-14
omf describe -- [mini-DoS]--> throttling/breakage


mobalt:
Asking GitHub for a description of 76 packages and 63 themes (139 separate total requests) with no sleep in between qualifies as a Denial-of-Service attack, albeit tiny. It's no surprise then that I get the first ~15 descriptions very quickly, then each subsequent one takes an additional 1 sec, then 3secs (throttling), then nothingness: The last packages show a null description (#212). Running `omf describe` a second time at any point within the hour returns nothing but empty strings.

The code responsible for sending 139 curl requests in parallel is located at **omf.packages.describe.fish** 
Sure enough, running curl by hand I get the 403 response below. Apparently github caps the api at 60 requests per hour? Hence why the last packages have guaranteed empty description.
`{
  "message": "API rate limit exceeded for 95.96.97.98. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}`

Two possible fixes:
1) Create a repository that keeps a json or csv of all plugin/theme repo metadata (#358) . If worried about making sure user always has the latest metadata when running `omf describe`, then just have a git pull right before returning results. As far as I know, github doesn't throttle pull requests from unauthenticated users.

2) Keep your curl code, but aim it at a single resource containing data for all repos. e.g.,
**`curl -s`https://api.github.com/users/oh-my-fish/repos**

coderstephen:
There was a lengthy discussion a while ago (albeit for a different reason) on a better way of packages storing metadata in a file in #196. Perhaps this should be visited again?

oranja:
Thank you for starting this discussion, @Mobalt. This is an important issue in my opinion.
There is no doubt that abusing github's API is something we want to avoid.

We've been talking about creating a central repository, along with a friendly website, for some time now, but it always gets pushed back in favor of other core features.
So approach (2), even though it's already a huge improvement over what we have now, would not yield comprehensive results just yet. We can spare a good amount of calls with a single query of **oh-my-fish/repos**, and perhaps that will be the solution in the short term, but it doesn't scale as long as we allow repositories outside of the *oh-my-fish* organization.
[...]

Issue #212, opened 2015-12-28
omf describe: Some plugins lack descriptions


simonbcn:
There are several plugins without any info with `omf describe`: 

```
$ omf describe composer
null
$ omf describe mix
$ omf describe stamp
$ omf describe vundle
null
```

unusedPhD:
i've started adding issues to the plugins that currently lack descriptions with recommendations.

oranja:
:+1:

btw, `stamp` points to a 'wa' repository which no longer exist. Does anyone has any idea what the plugin was for?
https://github.com/wa/pkg-stamp

derekstavis:
You can ignore the wa namespaced repositories. They were inherited from a brother framework, and most of them aren't compatible anymore.

bucaran:
@oranja That plugin was created mostly for testing purposes when I first started released Wahoo. You can safely delete it from the package list.

bpinto:
Thanks @unusedPhD that's a great idea.

sagebind:
This is now easily solvable by the organization maintainers and by the community with the merging of #427. Open a pull request on https://github.com/oh-my-fish/packages-main to add descriptions and other info on packages.

Issue #57, opened 2015-09-19
Plugin packaging


rominf:
I propose to reuse `git`'s infrastructure as much as possible:

```
$ omf info <plugin>
Name: <plugin name (echo repository_name | sed "s/plugin-//g")>
Version: <git tag; I think one number should be enough for a plugin version>
Installed: Yes
Status: up-to-date <is the tag last?>
Summary: <.git/description content>
Authors: <git commit authors (see http://stackoverflow.com/a/6482473/2108548)>
```

What do you think?

rominf:
Of course, README.md should contain only description. I'm against copy-pasing text like:
```
to install write:
...
to uninstall write:
...
```

I think that:
```
For more information on `oh-my-fish` visit ...
```
would be enough. This text is easy to append automatically by script to all plugins.

derekstavis:
I prefer using git tags instead of branches, this way plugins can be versioned using semver.

derekstavis:
But I definitely like `omf info` command and your proposed output!!

rominf:
> I'd prefer using git tags instead of branches, this way plugins can be versioned using semver.
You are right. I meant tags, not branches. I'll fix it.

bpinto:
I believe I didn't understand the proposal for the README file. Could you try explaining again?

rominf:
@bpinto 
Well. Printing markdown to a console output is not a very good idea.

I actually vote for self-documented code, so I think that after implementation of https://github.com/fish-shell/fish-shell/issues/478 we can just simply build README files from the help. So, for now, I would leave README files as they are.

BTW, it's possible to extract repository description from `.git/description` file.

bpinto:
I can see this working as it was proposed, might be interesting to see what current plugins would output.

jeremiejig:
See also `git describe` to generate a version, which relies on annotated tag.

derekstavis:
I usually use `git describe --tags --always HEAD`

bpinto:
@rominf are you working on a PR for this?

rominf:
@bpinto, no.

sagebind:
This will be marked as solved as part of the effort in #427.

sagebind:
#427 is merged. Packages can now be grouped, be more discoverable, and have metadata using the new package repository system built on top of Git. Closing as solved.

PR #299, opened 2016-04-13
doctor: check if Git remotes match db URLs

oranja:
For each installed package (plugin or theme), the doctor compares the remote URI (default remote configured for fetch on branch 'master') with the URI that's specified in the local package database.
If the _host_, _path_ or _repository name_ differs, a warning is issued. Protocols, logins and ports doesn't matter in this comparison.
As an example, `http://oranja@github.com/path/repo` in the database should have no problem with `git@github.com:path/repo.git` as the remote URI.

Warnings are also issued for installed packages that have no corresponding database entry. The exceptions to this rule are the built-in `omf` and `fish-spec` packages.

PR #427, opened 2016-10-23
New db system with metadata

sagebind:
This PR adds a major new change to Oh My Fish in how it handles its package database. Instead of bundling the database with the current framework version, this splits the database off into a separate repository that can be updated independently. There are a few motivations for doing so:

- Now that the stable update channel is available, users who wish to use stable versions will not get new or changed packages without upgrading to `master` or waiting for a new release.
- Similar to the above reason, if we want to eventually be in Homebrew (#3) or other package managers we will need to be able to get package updates without being able to update the core framework itself.
- Making changes to packages does not require waiting on Travis for testing, nor does it need protected branches. We can make the database repository a lot more open to commits without risk to stability of OMF.

As part of this change, this improves the database and package system in the following ways:

- Package entries are now property-like files containing key-value pairs:

  ```
  repository=https://github.com/oh-my-fish/plugin-await
  description=Wait last background job with a nice progress spinner
  ```

  This allows us to easily manage properties like descriptions without depending on the upstream package owners, while still allowing the maintainer to contribute and update the properties. This solves these issues: #212 #386. This also allows us to add more keys to the database in the future (license, package types?) without breaking compatibility.
- `omf submit` has been removed. It was confusing and didn't get much use, and would be even more confusing to use if the database is in a completely different Git repository.
- New `omf.db.query` function which can be used to perform searches on the database. The search syntax is not currently exposed to any user commands and only supports a few parameters, but this could be extended and improved further.
- Themes are no longer distinguished by `theme-` prefix on install, and instead are detected by looking for a `fish_prompt.fish` file. We should discuss if there remains any benefit to keeping packages and themes separate in the database.
- `omf describe` shows more detail on packages, similar to the proposal in #57. Can be easily extended to use more properties as they are added to the database.

Currently there is no direct way to update the database, and it is done so automatically during `omf install` and `omf update`. Do we need some sort of explicit `omf update-db`-type command?


lfiolhais:
This looks really awesome and is a much needed upgrade to OMF! 😄 One question, though. When I built the search functionality I couldn't find a good way to search themes or packages descriptions without blowing up Github's API. That was a feature that I ultimately decided against implementing. Now because this allows us to have a centralized and local location to execute searches on, this seems to be doable. That is my only pet peeve, I'd be glad to help.

derekstavis:
👏 👏 👏 👏 👏 👏 

Awesome ideas! Nice, beautiful code!!

I'm going through the commits and looking forward to see the branch upstream  🎉

derekstavis:
Some random thoughts:

* I see the repository is called `db`. This is a good opportunity of naming it properly. We could use something like `packages`, which is more meaningful than the legacy-inherited `db`.

* We need to version-match database to OMF version. This is needed because if we want to refactor the database interface in the future (eg. flattening the tree) we can lock previous OMF versions to previous databases versions.

sagebind:
While as-is the code mostly works, this is definitely a work-in-progress, so keep the suggestions coming!

@lfiolhais I didn't have time to add full-text searching, but that's something I definitely want to see be a core feature. Maybe I can merge the functions that read the properties and the one that searches into one function.

@derekstavis I'm not a big fan of the `db` name either, but I wasn't sure what to call it. I'm planning on extending this to support multiple package "lists" instead of hardcoding Oh My Fish's official one, kind of like Homebrew taps or APT repositories.

It might be tricky to version-lock the database interface if there are other ones out of our control. Hmm... perhaps a file containing the database info at the root, like minimum a minimum format version?

```fish
$ ls some/db/packages
...
$ cat some/db/info
name=some db
maintainer=??
db_format_version=2
```

I'm thinking of calling them either "(package) repositories", or "registries", or maybe something fun like "shoals".

sagebind:
@derekstavis To solve the version-lock issue (so that we don't just error saying DB version is incorrect), we could have an update-able list of repos to use, but include the Git branch/tag:

```fish
$ cat $OMF_CONFIG/lists
https://github.com/oh-my-fish/db master
```

If we need to save the current db format before changing it, legacy users could choose a specific branch/tag in the file to stay back:

```fish
$ cat $OMF_CONFIG/lists
https://github.com/oh-my-fish/db legacy-v2
```

Or alternatively have no `master` and just match branches to DB versions:

```fish
$ cat $OMF_CONFIG/lists
https://github.com/oh-my-fish/db version-two
```

derekstavis:
We could supercharge your idea by supporting lists on both `OMF_PATH` and `OMF_CONFIG`. The ones in `OMF_PATH` are the official ones, which are updated together with the framework, and `OMF_CONFIG` are the user's custom lists 👍 

Also, all valid package repositories could have a "manifest" file which adds some metadata about the repository, like name and description, and also framework version, so we could enable/disable repositories when detected version is incompatible.

sagebind:
@derekstavis I think we should avoid the repository-wide manifest file for now. It would add a lot of complexity since we'd have to clone the repository to tmp, read the manifest, verify format version, then move / replace the clone into the index. We can always add it later if it is refactored that much and call the manifest-less format as format version 1. I think bundling an always-working URL in `$OMF_PATH/repositories` will cover ourselves at least from problems with the official repository.

sagebind:
Ready for review. The `{packages,themes}` directory structure is maintained, and there is no repository manifest. Repositories can be specified (with branches) in `$OMF_CONFIG/repositories`, and is also specified in `$OMF_PATH/repositories`.

sagebind:
I didn't add a whole lot to the docs, I think we can add that as part of a separate PR (as well as a few other docs improvements I'd like to make). Primarily, putting command usage details into `omf help` instead of the docs, and put more high-level info in the docs.

`omf repo` calls `omf.index.repositories` directly instead of a `omf.cli.repositories` function. I'd like to start making the OMF functions more orthagonal.

derekstavis:
I'd just like to say that I don't like using awk too much. Code written in awk isn't very clear and this also imposes a new external dependency. I'd like to see how could we deal using plain fish code.

pkg/omf/functions/index/omf.index.query.fish
```diff
@@ -0,0 +1,82 @@
...
+    return 1
+  end
+
+  # Perform a text search if any textual terms were given.
+  if test -n "$q_name" -o -n "$q_text"
+    set results (command awk '
+      BEGIN {
+        FS = "[ \t]*=[ \t]*";
+        q_name = tolower(ENVIRON["q_name"]);
+        q_text = tolower(ENVIRON["q_text"]);
```
sagebind: Good tip, I'll try that.

sagebind:
@derekstavis I get that, but there's no way the full-text searching and property file parsing would be nearly as fast or maintainable in pure Fish as in AWK. Besides, AWK is portable since it's part of the POSIX standard.

pkg/omf/functions/index/omf.index.query.fish
derekstavis: Ah, true. Sometimes I forget that fish is awesome and that stuff like `-lx` is possible!

sagebind: `-v` wasn't working how I needed at some point so I used environment variables instead. I don't remember what the problem was but seems to work fine now. `-v` is more explicit.

derekstavis:
I believe in your motivations and trust your design decisions. So let's try to do some awk 👍 

I'm progressing in testing the branch locally, everything went smooth until now on macOS. I will give feedback if I find any issue in my environments. Also, I can give some Linux testing just next week, where I will be in touch with my limited embedded environments and my main Linux machine.

sagebind:
I've tested it in the `ohmyfish/ohmyfish` Docker image is all so far. It is a good way to test Linux since it is kind of the most minimal Linux you can get.

sagebind:
If we do want to merge packages and themes someday, I see no issue with doing that now in the repository. Instead of `themes/bobthefish`, just have `packages/bobthefish` with a `type` property:

```
$ cat packages/bobthefish
type = theme
repository = https://github.com/oh-my-fish/theme-bobthefish
```

Maybe a slight performance impact, but not enough to matter. I'd do this if it was a possibility to merge types in the future.

pkg/omf/functions/cli/omf.cli.describe.fish
```diff
@@ -1,4 +1,12 @@
-function omf.cli.describe -a name
-  omf.packages.describe $name
-  return 0
+function omf.cli.describe -a package -d 'Show information about a package'
+  if set -l props (omf.index.stat $package description repository maintainer)
+    echo "Package: $package"
+    echo "Description: $props[1]"
+    echo "Repository: $props[2]"
+    echo "Maintainer: $props[3]"
```
oranja: What happens if one of the keys are missing in the index entry?
I wouldn't bother with it now, because for now we're the only ones maintaining an index, and it's not so hard to keep a few keys for each package.
Just saying... builtin associative arrays in fish would fix this in a heartbeat (and will probably let some PowerShell ugliness into our beloved fish).

sagebind: Don't worry, the `omf.index.stat` command will return a blank line for property names not found. This works when any field is missing and just shows an empty string as the value.

pkg/omf/functions/cli/omf.cli.describe.fish
```diff
@@ -1,4 +1,12 @@
...
+
+    return 0
+  end
+
+  return 1
```
sagebind: Good idea, I'll do that.

pkg/omf/functions/cli/omf.cli.help.fish
```diff
@@ -169,23 +170,23 @@ Update Oh My Fish.
...
+  "(omf::em)"l"(omf::off)"ist          List local packages.
+  "(omf::em)"d"(omf::off)"escribe      Get information about what packages do.
+  "(omf::em)"i"(omf::off)"nstall       Install one or more packages.
+  "(omf::em)"t"(omf::off)"heme         List / Use themes.
+  "(omf::em)"r"(omf::off)"emove        Remove a theme or package.
+  "(omf::em)"u"(omf::off)"pdate        Update Oh My Fish.
+  "(omf::em)"c"(omf::off)"d            Change directory to plugin/theme directory.
+  "(omf::em)"n"(omf::off)"ew           Create a new package from a template.
+  "(omf::em)"s"(omf::off)"earch        Search for a package or theme.
+  "(omf::em)"repositories"(omf::off)"  Manage package repositories.
```
sagebind: Good plan. I'll change it.

pkg/omf/functions/cli/omf.cli.search.fish
```diff
@@ -1,26 +1,34 @@
-function omf.cli.search -d "CLI parser for the search command"
-
+function omf.cli.search -d 'Searches all available packages'
   switch (count $argv);
     case 1;
-      omf.search.pkg $argv[1]
+      echo (omf::under)"Packages"(omf::off)
+      __omf.cli.search.output type=package text=$argv[1]
```
sagebind: Mostly parse-ability. I'll probably change it to `--key=value`.

pkg/omf/functions/index/omf.index.query.fish
```diff
@@ -0,0 +1,102 @@
+function omf.index.query -d 'Query packages in the index'
+  set -q XDG_CACHE_HOME
+    and set -l index_path "$XDG_CACHE_HOME/omf"
+    or set -l index_path "$HOME/.cache/omf"
```
sagebind: Smart thinking. I'll do that.

pkg/omf/functions/index/omf.index.repositories.fish
```diff
@@ -0,0 +1,60 @@
...
+  switch $argv[1]
+    case help --help -h
+      omf help repositories
+      return
+
+    case list
+      for file in {$OMF_PATH,$OMF_CONFIG}/repositories
+        if test -f $file
+          read -z -l contents < $file
+          printf "$contents"
```
sagebind: I was thinking it was going to be more efficient. I just tried a `read` vs `cat` benchmark with 10,000 iterations and saw no difference on my machine, so maybe it isn't...

pkg/omf/functions/index/omf.index.repositories.fish
```diff
@@ -0,0 +1,60 @@
...
+      if not set -q argv[2]
+        echo "Error: url not specified" >&2
+        return 1
+      end
+
+      if not set -q argv[3]
+        set argv[3] master
+      end
+
+      echo "$argv[2] $argv[3]" >> $OMF_CONFIG/repositories
```
sagebind: Yeah good variable names are always better. I guess I got lazy there.

pkg/omf/functions/packages/omf.packages.install.fish
```diff
@@ -11,44 +11,50 @@ function __omf.packages.install.error.already
...
+    return $OMF_UNKNOWN_ERR
+  end
 
-    # Run the install hook.
-    if not omf.packages.run_hook $OMF_PATH/$parent_path/$name install
-      __omf.packages.install.error "$install_type $name"
-      return $OMF_UNKNOWN_ERR
-    end
+  # Check if the package is a theme. If it is, move it to the themes directory.
+  if test -f $install_dir/fish_prompt.fish
```
sagebind: Well, I don't think the first is achievable at all. We still have to consider arbitrarily-given Git repositories to be installed, (e.g. `omf install http://github.com/vendor/plugin`), which we can never guarantee to have metadata for.

I think the second is helpful because it leaves the name of the repository for themes up to the author, instead of forcing them to follow a specific naming pattern just to get it to work.

It sort of makes sense in this PR, especially if we do what I mentioned in [this comment](#issuecomment-271017252). If you want I can move it into a separate PR. Your comment I think pushed me over the edge of decision, so I'll implement the above comment and go from there.

pkg/omf/functions/index/omf.index.stat.fish
```diff
@@ -0,0 +1,55 @@
...
+        }
+      }
+    }
+
+    END {
+      for (i = 1; i <= count; i++) {
+        print values[i];
+      }
+    }
+  ' - $properties < $package_file
```
sagebind: It's a BSD thing... the BSD AWK implementation is garbage. I don't recall what the issue is right now, but I _believe_ it was that the BSD AWK refuses to read from STDIN unless `-` is a file argument, or 0 arguments are passed.

pkg/omf/functions/cli/omf.cli.search.fish
```diff
@@ -1,26 +1,34 @@
-function omf.cli.search -d "CLI parser for the search command"
-
+function omf.cli.search -d 'Searches all available packages'
   switch (count $argv);
     case 1;
-      omf.search.pkg $argv[1]
+      echo (omf::under)"Packages"(omf::off)
```
sagebind: I see what you're saying, and I think I agree. I'll change things to say "plugin" when talking about a "non-theme package". Note that "omf search" called plugins "packages" already for a little while now (but omf.index.query doing so _is_ my fault).

pkg/omf/functions/index/omf.index.query.fish
```diff
@@ -0,0 +1,102 @@
...
+  # If there are no packages, there is nothing to search.
+  if not set -q packages[1]
+    return 1
+  end
+
+  # Perform a text search if any textual terms were given.
+  if test -n "$q_name" -o -n "$q_text"
+    set results (command awk -v "q_name=$q_name" -v "q_text=$q_text" '
+      function flush() {
+        if (package && name_matches && text_matches) {
```
sagebind: When you put it that way, I don't see much of a need for the `name` filter...

Currently most of these can be done with the function though:

1. `omf.index.query --text=pattern` searches the name and any property value.
2. Restricting to specific property can't be done.
3. `omf.index.query --name=pattern`
4. `omf.index.query --name=pattern1 --text=pattern2`

Note that the default patterns for both `text` and `name` is `''`, which matches everything.

pkg/omf/functions/index/omf.index.update.fish
```diff
@@ -0,0 +1,57 @@
...
+    set repositories (command awk '
+      {
+        dir = $1 "_" $2;
+        gsub(/[ \/:@]/, "_", dir);
+        if (!visited[dir]++) {
+          print $1 "\n" $2 "\n" dir;
+        }
+      }
+    ' $lists)
+  end
```
sagebind: We could add that later when #299 is merged (soon maybe?). It'll still work OK if there are duplicates, even though it would be inconvenient.

pkg/omf/functions/index/omf.index.stat.fish
oranja: Ohh. I'm sorry for your pain.

pkg/omf/functions/packages/omf.packages.install.fish
oranja: You are totally right about the packages outside the index. For some reason i had in mind that we're also adding a manifest inside each package repository. Though it might make more sense to declare the type inside the package and not in the index, but that's not what this PR is about.
Having said that, I still think it might be better to split this to a separate PR, even just so it's not forgotten when we (probably you...) collect the changes for the next release announcement. Same goes to the removal of `omf submit`, though I'm not sure if anyone cares about it at all.

pkg/omf/functions/index/omf.index.query.fish
oranja: My bad. I was distracted and misunderstood the role of the 'text' keyword.
Everything makes more sense now. Scratch my comment here.

pkg/omf/functions/index/omf.index.repositories.fish
```diff
@@ -0,0 +1,67 @@
...
+  if not set -q argv[1]
+    set argv[1] list
+  end
+
+  switch $argv[1]
+    case help --help -h
+      omf help repositories
+      return
+
+    case list
```
derekstavis: maybe also `case list ls`?

PR #448, opened 2016-12-03
Add plugin for docker completion

wfscheper:
Repo is currentlyl at https://github.com/wfscheper/plugin-docker


bobthecow:
For completion-only plugins, we strongly recommend trying to get them merged into fish core. The core team is very willing to accept completions :)

PR #456, opened 2016-12-17
Print some useful diagnostic info in `omf doctor`

sagebind:
Since there are some common things we need to know when troubleshooting OMF, this adds some relevant diagnostic information so far as output to the `omf doctor` command. We can then recommend users simply paste the output of `omf doctor` as long as it is accessible, or literal `omf.doctor` otherwise.

This does not aid troubleshooting the installation process itself. It might be useful to separate the doctor into its own script that can be downloaded and used separately.


bobthecow:
:+1:

Issue #454, opened 2016-12-15
repository 'default' does not exist


mmahalwy:
I cannot get around this error:
```
Setting up Oh My Fish configuration...
Installing package default
fatal: repository 'default' does not exist
Could not install package default. 1^
fish: Unknown command '2'
~/.local/share/omf/pkg/omf/functions/packages/omf.packages.install.fish (line 1): echo (omf::err)"Could not install $argv."(omf::off) 1^&2
                                                                                                                                         ^
in function '__omf.packages.install.error'
	called on line 49 of file ~/.local/share/omf/pkg/omf/functions/packages/omf.packages.install.fish
	with parameter list 'package default'

in function 'omf.packages.install'
	called on line 22 of file ~/.local/share/omf/pkg/omf/functions/bundle/omf.bundle.install.fish
	with parameter list 'default'

in function 'omf.bundle.install'
	called on line 6 of file ~/.local/share/omf/pkg/omf/functions/cli/omf.cli.install.fish

in function 'omf.cli.install'
	called on line 47 of file ~/.local/share/omf/pkg/omf/functions/omf.fish

in function 'omf'
```

derekstavis:
Hi @mmahalwy, thanks for your report!

What system and fish version are you using?

Could you also please check if the command below returns something?

```
git config --global core.autocrlf
```

mmahalwy:
fish: fish, version 2.4.0-231-gd885f00
OS: macOS Sierra
Git: git version 2.11.0

Command
```
git config --global core.autocrlf
```
Does not return anything

oranja:
Hi @mmahalwy,
Two more troubleshooting questions:

1. Do you sometimes have internet connection issues?
2. Can you please paste the output from
`cat ~/.local/share/omf/db/themes/default`
?

For some reason the installer assumed that *default* is a name of a plugin and not a theme. From there it looked for the URL in the wrong place and obviously failed to clone it.

Note to ourselves:
This is not the root of the problem here, but `1^&2` is not a valid syntax in fish 2.4.
Maybe it was sometime before and it changed. Now this should be `1>&2`.

sagebind:
@oranja Looks like the redirection changed in 2.4: https://github.com/fish-shell/fish-shell/issues/1873

This should be fixed in the install code.

mmahalwy:
@oranja 
```
https://github.com/oh-my-fish/theme-default
```

oranja:
[...]

PR #459, opened 2016-12-17
Normalize redirection syntax

sagebind:
Fixes a problem with redirection syntax for Fish >=2.4.0 observed in #454. All redirections are normalized to the correct syntax.

Also bump Fish version in Docker image for better catching of these issues.


oranja:
You beat me to it.
💯 👍

PR #468, opened 2017-01-04
Prepare release 4

sagebind:


oranja:
LGTM, but why not merge the Ukrainian translation (#464) and add it to v4?

sagebind:
@oranja Sure, I can do that.

Issue #489, opened 2017-02-14
http://get.oh-my.fish/ is blank


NinoFloris:
Might want to take a look as the installation does not work

sagebind:
@NinoFloris Tried it just now and it works. If you're using cURL, be sure to enable following redirects with `-L`:

```
curl -L http://get.oh-my.fish
```

NinoFloris:
Nope was also blank in browser, works here now again as well. Thanks for your quick response, must have been a fluke.

Issue #482, opened 2017-02-04
[Discussion] Downloading the install script over an HTTP connection is a bad idea


N0ps32:
The user is executing untrusted code that has been transferred over an insecure connection. Well technically only the redirect is insecure but since it's running over HTTP it's extremely easy to execute a MITM attack here and potentially execute harmful code by changing the redirect destination.

I think a good example of how this should be done is composer: https://getcomposer.org/download/

The script is:

- Always transferred over an HTTPS connection
- Verified with a hash once it's downloaded

Considering the "get.oh-my.fish" domain only returns a redirect anyway wouldn't it be better to just directly include the secure githubusercontent URL? Or at least verify the script integrity with a hash before executing it. I know the command won't look as "sexy" and short as it does now but it would really help improve security.

(or just get a free SSL certificate and migrate to HTTPS)

oranja:
We have a SSL certificate for _oh-my.fish_, but when I just set the domain up I opted to let the DNS provider handle the simple redirect. My personal unfamiliarity with `nginx` and infrastructure management in general sounded like an equally dangerous risk factor.
I agree that the insecure HTTP redirect should be gone.

As an added value, we then would be able to count downloads and also might get rid of the recent nginx errors that we are getting from the DNS and redirect provider.

The hash verification would also be a great addition, and the more streamlined (mandatory but not interrupting), the merrier.

My only caveat is the "extremely easy" part. I agree that it's extremely possible, but what makes you say that it's extremely easy to implement in practice? I'm asking so that everyone would get a better understanding of the threat model.

sagebind:
Solved by #490 and #504.

Issue #478, opened 2017-01-29
install with curl not working


oprogramador:
When I run `curl -L http://get.oh-my.fish | fish` (as in README), I obtain:

```
fish: Encountered redirection when expecting a command name. Fish does not allow a redirection operation before a command.
Standard input: <html>
                ^
fish: Encountered redirection when expecting a command name. Fish does not allow a redirection operation before a command.
Standard input: <head><title>400 Bad Request</title></head>
                ^
fish: Encountered redirection when expecting a command name. Fish does not allow a redirection operation before a command.
Standard input: <body bgcolor="white">
                ^
fish: Encountered redirection when expecting a command name. Fish does not allow a redirection operation before a command.
Standard input: <center><h1>400 Bad Request</h1></center>
                ^
fish: Encountered redirection when expecting a command name. Fish does not allow a redirection operation before a command.
Standard input: <hr><center>nginx</center>
                ^
fish: Encountered redirection when expecting a command name. Fish does not allow a redirection operation before a command.
Standard input: </body>
                ^
fish: Encountered redirection when expecting a command name. Fish does not allow a redirection operation before a command.
Standard input: </html>
```

oprogramador:
However not always

oranja:
Our domain provider seems to be having some issues lately. I saw the **400 Bad Request** page again just a few seconds ago.

I will let them know about these instabilities, but for now, for anyone that's interested, know that http://get.oh-my.fish is simply a shortcut (redirection) to:
https://raw.githubusercontent.com/oh-my-fish/oh-my-fish/master/bin/install
So in "time of trouble" like these, you can install with:
```shell
curl -L https://raw.githubusercontent.com/oh-my-fish/oh-my-fish/master/bin/install | fish
```
It's a little longer to type, but as a copy-paste one-liner it's all the same. ;)

sagebind:
This seems to be "answered", so closing the issue. We might have DNS problems in the future or address it. The team will handle this internally.

PR #490, opened 2017-02-17
Get oh-my-fish installer from a secure connection.

oranja:
Use HTTPS schema instead of cleartext HTTP.
SSL certificate provided by letsencrypt.com, installed on private nginx server for redirection.

Moving to the private server should solve the issue with NameCheap's URL redirection service that didn't behave so well lately (see: #478, #489).
The move to HTTPS is a reasonable and meaningful step towards a safer environment for OMF users (see discussion in #482).

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,13 @@ Oh My Fish provides core infrastructure to allow you to install packages which e
 You can get started right away with the default setup by running this in your terminal:
 
 ```fish
-curl -L http://get.oh-my.fish | fish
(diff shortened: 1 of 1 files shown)


consultantRR:
The SSL certificate has expired:

``` bash
$ echo | openssl s_client -servername get.oh-my.fish -connect get.oh-my.fish:443 2>/dev/null | openssl x509 -noout -dates
notBefore=Feb 17 20:21:00 2017 GMT
notAfter=May 18 20:21:00 2017 GMT
```

PR #502, opened 2017-04-24
Refactor command execution and help

sagebind:
- Copy and improve help text, and convert it to AsciiDoc files
- Evaluate commands dynamically

Why AsciiDoc? It supports definition lists, and Markdown doesn't. The source AsciiDoc formatting also required much less work to transform into something that looks good in the terminal.

If we could handle aliases in some way, we could remove the switch statement altogether and compute the list of available commands by listing the files in `$OMF_PATH/pkg/omf/functions/cli`.


derekstavis:
Just checkouted the branch. Awesome work!! Looks **very** beautiful!!

I don't know if was intentional, but I see `::` appended to every command or switch in `omf help`:

screen shot 2017-06-25 at 18 24 38

sagebind:
@derekstavis Oh boy, that's not supposed to happen! I'll look into the AsciiDoc parser and see if there's a bug.

sagebind:
Darn BSD incompatibilities... that's what I get for not testing on OS X as well as Linux.

derekstavis:
It works now!! Getting back to review!

derekstavis:
Wow, it works awesomely well. Extra points for the comments in the `sed` chain 😛 

LGTM :shipit:

sagebind:
Sweet; merging into master!

PR #505, opened 2017-04-27
Limit search to name+description

sagebind:
Full text search outside of name and description seems to make the search _less_ useful. For example, searching for "git" returns every possible package, because they all have "github" in the repo URL. Similarly searching for other fields seems to reduce the usefulness of the results.

This tightens up the search by applying full text search queries to only the package name and description.


bobthecow:
:shipit:

PR #515, opened 2017-05-24
docs: fix `pure` screenshot

rafaelrinaldi:


sagebind:
I think we've had this issue before. The problem is that `Themes.md` is automatically generated using the `tools/generate-themes-doc.fish` script, which we run every once and a while. Every time we do, the screenshot for Pure breaks, because that project's README has a relative image URL instead of an absolute one.

PR #520, opened 2017-06-13
omf search update in readme. --pkg to -p 

Necmttn:
a small change in the readme file.


scorphus:
Yeah, good catch!

Let me just squash those two commits into one, it kinda bugs me having a merge commit as part of a pull request.

PR #553, opened 2017-09-06
Update README.md to reflect length of list.

duner:
Hi! I was about to install oh-my-fish for the first time and saw this discrepancy in the README. The introduction to the list had four items, but the list itself has five items.


sagebind:
@duner Thanks for fixing that! Looks like this is no longer an issue though since #546 is merged.

PR #563, opened 2017-10-05
Do not create symlinks to `fish_prompt.fish` files

### LINKED ISSUE #562: fish_config won't save settings
Please read this [issue](https://github.com/fish-shell/fish-shell/issues/4357#event-1238427501)
It is solved but it seems to be an omf bug

oranja:
Autoloading the theme directory should be enough.

---
This fixes #562.
Additional testing and review welcome as always.

### DIFF
--- a/pkg/omf/functions/themes/omf.theme.set.fish
+++ b/pkg/omf/functions/themes/omf.theme.set.fish
@@ -27,13 +27,6 @@ function omf.theme.set -a target_theme
   set -l theme_path {$OMF_CONFIG,$OMF_PATH}/themes*/$target_theme{,/functions}
   autoload $theme_path
 
-  # Find target theme's fish_prompt and link to user function path
-  for path in {$OMF_CONFIG,$OMF_PATH}/themes/$target_theme/$prompt_filename
-    if test -e $path
-      ln -sf $path $user_functions_path/$prompt_filename; and break
-    end
-  end
-
   # Reload fish key bindings if reload is available and needed
   functions -q __fish_reload_key_bindings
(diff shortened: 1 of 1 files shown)


sagebind:
@oranja I've had a fix like this in the `wip/require-theme-no-fs` branch for a very long time, but the problem with doing this is that it breaks theme switching.

Setting the theme using `fish_config` will copy the desired `fish_prompt.fish` file into the user's `functions` directory. Then later, running `omf theme <whatever>` will do nothing, since it is always going to be overridden with the `fish_prompt.fish` file created earlier.

SkylerWittman:
This is quite old but I'm pretty sure I'm affected by this. I'm using https://github.com/oh-my-fish/theme-bobthefish and everytime I close and re-open my shell (iTerm2), it's resetting my colour scheme in bobthefish back to the default one even though I set the colour scheme globally to be the one I chose with `set -x`.

bobthecow:
@SkylerWittman (and anyone else who runs into this) removing the symlink or copied file should fix it (though you might need to re-open fish after):

```
rm $user_functions_path/fish_prompt.fish
```

@sagebind Should we remove the symlink as in this PR, then update theme setting (and `omf doctor`?) to warn if there's a prompt at `$user_functions_path/fish_prompt.fish`?

hedy06:
Does that mean I won't be able to install additional prompts from omf, and then switch back to default prompt, without having to remove the theme and the sym link?

oranja:
This is too old now and there are some other (better) efforts to resolve the theme switching issue. Closing this one.

PR #572, opened 2017-10-31
Add capability to list functions of a package

NikolaiMagnussen:


NikolaiMagnussen:
Sorry for being so slow. Forgot this a bit:
1. (Will be) fixed by looping over the list returned by the grep call - just display all can be a good idea.
2.
    a. My idea was for this to list functions that can be run at any time, and I assume  the ones in init.fish will only run once - on startup, and I don't know if that should be considered in-scope(?)
    b. This was primarily for providing an easy way of listing the functions that are exported by an omf package, so I don't know if I would consider it in-scope to cover themes. Disclaimer: I don't know anything about themes and how they are built.
3. I did not know if that function, and will use it instead of my hacky path substitution

PR #603, opened 2018-03-30
use path queries compatible with old fish

### LINKED ISSUE #602: install failed
```
bin/install --offline  
type: invalid option -- 'q'  
Install aborted: Missing required command: awk
```
And
```
apt list --installed |grep awk
grep: warning: GREP_OPTIONS is deprecated; please use an alias or script

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

gawk/trusty,trusty,now 1:4.0.1+dfsg-2.1ubuntu2 armhf [installed]
mawk/trusty,trusty,now 1.3.3-17ubuntu2 armhf [installed]
original-awk/trusty,trusty,now 2012-12-20-1 armhf [installed]
```
How should I do?

sagebind:
Fixes #602. Previously on old Fish versions not supporting `type -q`, the installer would fail to check for required commands and print obscure and misleading errors. Since we can't move checking the Fish version before we check for commands (checking the version itself needs `sort` and `head`), make searching for commands compatible with older Fish versions.


bobthecow:
🆗

PR #618, opened 2018-06-08
Use posix stderr redirection (`2>` instead of `^`)

### LINKED ISSUE #609: Remove caret (^)
The caret operator as been remove in the master branch of fish-shell
so the installation script isn't working anymore

oranja:
Fish is deprecating `^` as a shortcut for stderr redirection (fish-shell/fish-shell/issues/4394).
Replace with the more common and compatible `2>`.

Fixes #609.

We should try to propagate this to packages too..


bobthecow:
>We should try to propagate this to packages too..

:+1: I’ll do a bulk PR to get all the packages up to date.

PR #642, opened 2018-09-30
Updated readme

sathiz1993:
Updated readme for display master branch name in shell


bobthecow:
This is a theme specific configuration, and should be documented by the themes, not by OMF. For example, See the configuration section of bobthefish’s readme: https://github.com/oh-my-fish/theme-bobthefish/blob/master/README.md

PR #651, opened 2018-10-25
Check whether we are inside the worktree

krobelus:
Some git commands require to be run from inside the worktree (as opposed to the git dir, although it's usually in worktree/.git). This commit adds a function git_is_worktree to check this. It is used for the commands that need the worktree instead of git_is_repo.

An alternative solution might have been to find the git worktree in the parent of the git directory, but this doesn't work for all cases. Generally it's impossible to detect the location of the worktree (plus it's not unique).


faho (approved):
This LGTM, the merge conflict should be quite simple.

bobthecow (approved):
LGTM

scorphus (approved):
Thanks, @krobelus!

scorphus:
If I haven't merged this when TravisCI finally gives it [a chance to build](https://www.traviscistatus.com/#system-metrics), feel free to act. I'd suggest you to use the "Squash and merge" button.

faho:
> If I haven't merged this when TravisCI finally gives it a chance to build

Side note: You need to move away from Travis. See https://github.com/fish-shell/fish-shell/issues/7447

scorphus:
Yep. I've done this for several other projects. Travis was so nice since the old days. But it was sold and since then it's been focusing more on enterprise contracts. It's GitHub actions FTW.

bobthecow:
> You need to move away from Travis.

Yes.

PR #652, opened 2018-10-25
Fix error when inside a bare git repository

krobelus:


lib/git/git_is_repo.fish
```diff
@@ -1,3 +1,7 @@
 function git_is_repo -d "Check if directory is a repository"
-  test -d .git; or command git rev-parse --git-dir >/dev/null ^/dev/null
+  test -d .git
+  or begin
+    command git rev-parse --git-dir >/dev/null ^/dev/null
```
faho: This would have to be changed to `2>` instead of `^`.

Also it's possible to only invoke `git rev-parse` once:

```fish
set -l info (git rev-parse --git-dir --is-bare-repository 2>/dev/null)
and test $info[2] = false
```

krobelus: good points, done
I actually didn't know that `set` propagates the failure, totally makes sense this way.
thanks!

faho (approved):
LGTM!

lib/git/git_is_repo.fish
```diff
@@ -1,3 +1,7 @@
 function git_is_repo -d "Check if directory is a repository"
-  test -d .git; or command git rev-parse --git-dir >/dev/null 2> /dev/null
+  test -d .git
+  or begin
+    set -l info (git rev-parse --git-dir --is-bare-repository 2>/dev/null)
```
bobthecow, line 4: this should still be `command git rev-parse`. it's pretty common to override `git` to provide additional functionality or default options, so let's be defensive about this.

bobthecow (approved):

scorphus:
Thank you, @krobelus! 🐠🐚

PR #655, opened 2018-11-08
Fix screenshot in themes preview

psliwka:
The link to image was broken.

Piggyback: align table cells.

### DIFF
--- a/docs/Themes.md
+++ b/docs/Themes.md
@@ -292,11 +292,11 @@ set -g theme_project_dir_length 1
 
 **Color scheme options**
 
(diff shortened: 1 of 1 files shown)


bobthecow:
Hey, thank you for noticing :) 

The Themes doc is automatically generated from all the READMEs for all the themes, so this is actually a bug in the theme preview generation script, probably on this line:

https://github.com/oh-my-fish/oh-my-fish/blob/master/tools/generate-themes-doc.fish#L18

The script adds namespaces to all the markdown named links, so that there aren’t collisions between themes after merging their readmes.

scorphus:
This issue is now fixed. Thanks for your contribution.

Issue #665, opened 2019-01-04
Conf.d might be loaded multiple times


thalesmello:
The [for loop to source conf_path](https://github.com/oh-my-fish/oh-my-fish/blob/master/lib/require.fish#L77) is inside of the [for loop to source init_path](https://github.com/oh-my-fish/oh-my-fish/blob/master/lib/require.fish#L48).

If there are any packages that use the `conf.d` folder, they might be loaded multiple, they might be sourced multiple times.

thalesmello:
@derekstavis you might want to take a look at this

PR #666, opened 2019-01-04
Put conf_file loop outside of init_file loop

thalesmello:
Addreses https://github.com/oh-my-fish/oh-my-fish/issues/665

### DIFF
--- a/lib/require.fish
+++ b/lib/require.fish
@@ -74,14 +74,14 @@ function require
 
     source $init $path
 
-    for conf in $conf_path
-      source $conf
-    end
-
     emit init_$package $path
 
     set -g omf_init_path $omf_init_path $path
     emit perf:timer:finish $init
+  end
+
+  for conf in $conf_path
+    source $conf
   end
 
   return 0


derekstavis:
Thanks for taking the time to debug and fix this bug!

The loop was done on purpose. As you can see, globs are used thoroughly in `require` function, and it is used mainly because it is a performant way of checking for the existence of one or more files without forking to a subprocess or calling `test` (which is prohibitive in the init code path).

The `$init_path` variable in this case is populated with all `init.fish` files present in package directory, which should return one and only one file (yes, prepending anything after `.fish` will give bogus stuff). Further in the code, if the given `init.fish` file was already sourced, we skip loading again, what would avoid reloading `conf.d` twice.

I noticed that this PR fixes a bug though, that I didn't notice when adding `conf.d` loading! Currently `conf.d` is only being loaded when `init.fish` is present, which is not what we wanted to achieve.

I'm not gonna block the merge as this is fixing a real bug, but I will leave some concerns that we might want to address in the future:

1. The mechanism that will not source `init.fish` (or init files, which includes `conf.d`) is not effective for `conf.d` anymore (https://github.com/oh-my-fish/oh-my-fish/blob/master/lib/require.fish#L55)
2. The benchmarking events will not take `conf.d` into account anymore (https://github.com/oh-my-fish/oh-my-fish/blob/master/lib/require.fish#L84)

PR #670, opened 2019-01-04
Remove %-substitutions from installer for Fish 3

sagebind:
Use the new fish_exit event instead of --on-process-exit %self. Include polyfill for older versions.

Note that this cannot be merged yet, as the new fish_exit event provides no way to get the status code.

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -8,11 +8,9 @@ set -q OMF_REPO_BRANCH; or set OMF_REPO_BRANCH "master"
 
 # fish_exit polyfill for Fish versions < 3.
 if not set -q fish_pid
-  eval '
-    function fish_exit_polyfill --on-process-exit %self
-        emit fish_exit $argv[3]
-    end
-  '
+  function fish_exit_polyfill --on-process-exit %self
+    emit fish_exit $argv[3]
+  end
 end
 
 
--- a/bin/install
+++ b/bin/install
@@ -6,6 +6,16 @@ set -q OMF_REPO_URI;    or set OMF_REPO_URI "https://github.com/oh-my-fish/oh-my
 set -q OMF_REPO_BRANCH; or set OMF_REPO_BRANCH "master"
 
 
+# fish_exit polyfill for Fish versions < 3.
+if not set -q fish_pid
+  eval '
+    function fish_exit_polyfill --on-process-exit %self
+        emit fish_exit $argv[3]
+    end
+  '
+end
+
+
 function main
   # Set default settings
   set -q XDG_DATA_HOME
@@ -137,8 +147,8 @@ end
 
 
 # Add an exit hook to display a message if the installer aborts or errors.
-function on_exit -p %self
-  if not contains $argv[3] 0 2
+function on_exit --on-event fish_exit -a exit_status
+  if not contains $exit_status 0 2
     echo "
 Oh My Fish installation failed.
 


bin/install
```diff
@@ -6,6 +6,16 @@ set -q OMF_REPO_URI;    or set OMF_REPO_URI "https://github.com/oh-my-fish/oh-my
...
 
 
+# fish_exit polyfill for Fish versions < 3.
+if not set -q fish_pid
+  eval '
+    function fish_exit_polyfill --on-process-exit %self
+        emit fish_exit $argv[3]
+    end
+  '
+end
```
scorphus: How about ditching the super slow `eval` and leaving only the function declaration itself:
```fish
if not set -q fish_pid
  function fish_exit_polyfill --on-process-exit %self
    emit fish_exit $argv[3]
  end
end
```
And we get increased readability as a bonus 😉

sagebind: Well if %-substitution is deprecated, then when it is removed will the above yield a syntax error when the file is parsed? Or will it just be a string whose value is `"%self"`? I originally used `eval` so that the deprecated syntax is conditionally parsed.

bobthecow: replace `%self` with `%monkey` to test :) 

the `if not set` seems to prevent the syntax error:
screen shot 2019-01-07 at 8 20 39 am

scorphus:
BTW, AFAIK `%self` [is not deprecated](https://github.com/fish-shell/fish-shell/releases/tag/3.0.0). Or is it going to?

sagebind:
> BTW, AFAIK `%self` is not deprecated. Or is it going to be?

No, it is not listed as deprecated, and the comments on [the actual PR](https://github.com/fish-shell/fish-shell/issues/4230) seems to indicate we can rely on `%self` for a while at least. The actual deprecation appears to be the `--on-process-exit` switch:

> The `function --on-process-exit` switch will be removed in future ([#4700](https://github.com/fish-shell/fish-shell/issues/4700)). Use the `fish_exit` event instead: `function --on-event fish_exit`.

sagebind:
Now this is interesting, because `fish_exit` will only provide the exit code in Fish >= 3.1.0. How do we detect 3.0.x in an infallible way? We don't want to do anything tricky, because we want to register the exit handler as soon as possible so that errors are caught.

Basically the matrix is like this:

- Fish 2.x.x must use `--on-process-exit %self`
- Fish 3.0.x must use `--on-process-exit %self`
- Fish >= 3.1.0 should use `--on-event fish_exit`

bin/install
faho: 1. %self will stay for at least a while

2. There is no *syntax error* here - without process expansion, the `%self` would simply stay `%self`, and the option to `function` will only be evaluated once that's actually called, so even if `--on-process-exit` was removed it would still work if it was never called.

So yeah, there's no need for the `eval`.

bin/install
```diff
@@ -6,6 +6,16 @@ set -q OMF_REPO_URI;    or set OMF_REPO_URI "https://github.com/oh-my-fish/oh-my
 set -q OMF_REPO_BRANCH; or set OMF_REPO_BRANCH "master"
 
 
+# fish_exit polyfill for Fish versions < 3.
+if not set -q fish_pid
+  eval '
+    function fish_exit_polyfill --on-process-exit %self
+        emit fish_exit $argv[3]
+    end
+  '
```
scorphus, line 15:
```suggestion
  function fish_exit_polyfill --on-process-exit %self
    emit fish_exit $argv[3]
  end
```

scorphus (changes requested):
Suggesting that change just so it can be committed.

sagebind:
This can't be merged even with the suggestions, because the `fish_exit` event did not include the exit code in Fish 3.0.x. Since `%self` is still valid, I don't see a reason to do this at all right now. Trying to use `fish_exit` isn't worth the complexity.

Maybe if `%self` is removed in Fish 4 (or 3.x), then we can worry about this by first checking if `test %self = '%self'`.

scorphus:
> if %self is removed in Fish 4 (or 3.x), then we can worry about this

Yup. I thought the very same. This could be left aside or even closed.

PR #682, opened 2019-02-07
Remove occurrences of `?` as a glob

### LINKED ISSUE #644: Check uses of "?" for fish 3 compatibility
Howdy,

it appears OMF uses the "?" glob in a few places. Fish 3 deprecates that one and allows users to disable it via a "feature flag" (like the `^` stderr redirection).

Going through the ones I've found via `git grep '\?' | grep -v .md:`:

These should presumably just use `*` instead:

```fish
init.fish:  set -l key_bindings $OMF_CONFIG/key_binding?.fish \
init.fish:                      {$OMF_CONFIG,$OMF_PATH}/themes/$theme/key_binding?.fish
pkg/omf/functions/themes/omf.theme.set.fish:    and test (count {$OMF_CONFIG,$OMF_PATH}/key_binding?.fish) -gt 0
```

These read the part after the "=" into a variable immediately afterwards, and the variable defaults to empty, so the `?` can presumably simply be dropped:

```fish
pkg/omf/functions/index/omf.index.query.fish:      case '--name=?*'
pkg/omf/functions/index/omf.index.query.fish:      case '--text=?*'
```

I'd probably just drop the `-?` and `/?` way of saying "--help":

```fish
bin/install:      case --help -h '/?'
pkg/omf/functions/omf.fish:      case '-h' '--help' '-\?' '/\?'
```

If you must offer it, matching a literal `?` here is a bit more involved than I'd like it, but since it doesn't happen often.... This is the code we use:

```fish
    # A literal "?" for use in `case`.
    set -l q '\\?'
    if status test-feature qmark-noglob
        set q '?'
    end

    # ...
    case "$q"
```

This is just an error for an unknown option, so it's probably also okay to trigger on just a "-":

```
pkg/omf/functions/omf.fish:      case '-?*'
```

(Alternatively, catch `case -` before, and then use `case '-*'`)

scorphus:
Since Fish Shell 3.0 [1], `?` as a glob is deprecated. Fixes #644

  1. https://github.com/fish-shell/fish-shell/releases/tag/3.0.0

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -29,7 +29,7 @@ function main
   # Parse command args
   while set -q argv[1]
     switch "$argv[1]"
-      case --help -h '/?'
+      case --help -h
         echo "Usage: install [options]
   Install Oh My Fish
 
--- a/init.fish
+++ b/init.fish
@@ -33,9 +33,9 @@ function fish_user_key_bindings
     and read -l theme < $OMF_CONFIG/theme
     or set -l theme default
   # Prepare packages key bindings paths
-  set -l key_bindings $OMF_CONFIG/key_binding?.fish \
+  set -l key_bindings $OMF_CONFIG/key_binding*.fish \
                       {$OMF_CONFIG,$OMF_PATH}/pkg/*/key_bindings.fish \
-                      {$OMF_CONFIG,$OMF_PATH}/themes/$theme/key_binding?.fish
+                      {$OMF_CONFIG,$OMF_PATH}/themes*/$theme/key_binding.fish
   # Source all keybindings collected
   for file in $key_bindings
     source $file
--- a/init.fish
+++ b/init.fish
@@ -32,10 +32,11 @@ function fish_user_key_bindings
   test -f $OMF_CONFIG/theme
     and read -l theme < $OMF_CONFIG/theme
     or set -l theme default
+  test -e $OMF_CONFIG/key_bindings.fish;
+    and source $OMF_CONFIG/key_bindings.fish
   # Prepare packages key bindings paths
-  set -l key_bindings $OMF_CONFIG/key_binding*.fish \
-                      {$OMF_CONFIG,$OMF_PATH}/pkg/*/key_bindings.fish \
-                      {$OMF_CONFIG,$OMF_PATH}/themes*/$theme/key_binding.fish
+  set -l key_bindings {$OMF_CONFIG,$OMF_PATH}/pkg/*/key_bindings.fish \
+                      {$OMF_CONFIG,$OMF_PATH}/themes*/$theme/key_bindings.fish
   # Source all keybindings collected
   for file in $key_bindings
     source $file
--- a/pkg/omf/functions/index/omf.index.query.fish
+++ b/pkg/omf/functions/index/omf.index.query.fish
@@ -12,10 +12,10 @@ function omf.index.query -d 'Query packages in the index'
         set q_type plugin
       case '--type=theme'
         set q_type theme
-      case '--name=?*'
+      case '--name=*'
         set -l IFS '='
         echo "$arg" | read dummy q_name
-      case '--text=?*'
+      case '--text=*'
         set -l IFS '='
         echo "$arg" | read dummy q_text
       case '*'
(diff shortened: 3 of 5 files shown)


bin/install
```diff
@@ -29,7 +29,7 @@ function main
   # Parse command args
   while set -q argv[1]
     switch "$argv[1]"
-      case --help -h '/?'
+      case --help -h
```
bobthecow: these aren't a glob, but seems reasonable to remove all the `-?` and `/?` args anyway?

init.fish
```diff
@@ -33,9 +33,9 @@ function fish_user_key_bindings
     and read -l theme < $OMF_CONFIG/theme
     or set -l theme default
   # Prepare packages key bindings paths
-  set -l key_bindings $OMF_CONFIG/key_binding?.fish \
+  set -l key_bindings $OMF_CONFIG/key_binding*.fish \
                       {$OMF_CONFIG,$OMF_PATH}/pkg/*/key_bindings.fish \
-                      {$OMF_CONFIG,$OMF_PATH}/themes/$theme/key_binding?.fish
+                      {$OMF_CONFIG,$OMF_PATH}/themes/$theme/key_binding*.fish
```
bobthecow: Switching from `?` to `*` in these leads to a much greater chance of false positives (maybe someone has a `key_bindings_old.fish` file or something) so I don't love it. Maybe moving the `*` into the path somewhere is a better approach?

```
$OMF_CONFIG*/key_bindings.fish \
  {$OMF_CONFIG,$OMF_PATH}/pkg/*/key_bindings.fish \
  {$OMF_CONFIG,$OMF_PATH}/themes*/$theme/key_bindings.fish
```

bin/install
scorphus: Yup, not a glob. I think it's awkward and not very common. That's my humble opinion, though. And I guess you agree on that?

init.fish
scorphus: I didn't think of that, good catch! I agree that's a better approach and the result is the same. Thanks!

bin/install
sagebind: I vote just remove it. It is the Windows way of asking for help, but people running OMF on Windows these days are just using WSL or cygwin anyway.

init.fish
```diff
@@ -33,9 +33,9 @@ function fish_user_key_bindings
     and read -l theme < $OMF_CONFIG/theme
     or set -l theme default
   # Prepare packages key bindings paths
-  set -l key_bindings $OMF_CONFIG/key_binding?.fish \
+  set -l key_bindings $OMF_CONFIG/key_binding*.fish \
```
bobthecow: should we move the `*` on this or the `off.theme.set` at the bottom?

scorphus: I'm looking into that.

Also, it's `key_bindings.fish`, with a trailing `s`, at least for the plugins and themes that have it:

* [plugin-bang-bang](/oh-my-fish/plugin-bang-bang)
* [plugin-expand](/oh-my-fish/plugin-expand)
* [plugin-sudope](/oh-my-fish/plugin-sudope)
* [theme-budspencer](/oh-my-fish/theme-budspencer)
* [theme-dangerous](/oh-my-fish/theme-dangerous)

scorphus: Having `$OMF_CONFIG*/key_bindings.fish` leads to problem in case of `~/.config/omf_old`, for example.

scorphus: Maybe there's no way to make it part of `$key_bindings` like the others. What if we solve it with the following?

```fish
test -e $OMF_CONFIG/key_bindings.fish; and source $OMF_CONFIG/key_bindings.fish
```

scorphus:
I'll run a few more tests tomorrow, I'm too tired now.

scorphus:
Tests completed. Versions tested:
Fish 2.3.1 ✅
Fish 2.7.1 ✅
Fish 3.0.0 ✅
Fish 3.0.1 ✅

🐟 🐚 FTW

scorphus:
@bobthecow, thanks for your reviews. When you have the time, please, I'd really like to read your inputs regarding the last two changes (they will be autosquashed). Thanks!

PR #694, opened 2019-04-02
update `pure` description to reflect `2.x`

edouard-lopez:
Hi,
I'm a [core maintainer of `pure` prompt](https://github.com/rafaelrinaldi/pure/graphs/contributors).
I simply updated `pure` description with our current readme content.


edouard-lopez:
ping @bobthecow @sagebind

derekstavis:
We have a script that we should be running to update this page, but it's been due for a while. Merging  this anyway, since it's better to be up to date...

PR #706, opened 2019-06-04
lib/git/git_is_touched: Speed up

### LINKED ISSUE #624: Changing to Git directories takes a long time
Changing to Git directories takes long if the repository (2,1 GB) is big.

faho:
This used to use `git status --porcelain`, which by necessity needs to
check the entire repo for all kinds of changes, just to figure out if
there are any.

Instead, we now use git commands that can exit early.

In large repos, this can be faster by a factor of 15 or so.

Fixes #624.


scorphus:
Thanks, @faho! <3

PR #718, opened 2019-08-23
Call the rm binary directly

### LINKED ISSUE #717: omf remove may fail when rm is aliased to something else
Having `rm` aliased to

```
function rm
    echo 'This is not the command you are looking for. Use tp'
    false
end
```

breaks `omf remove` since it calls the alias and not the binary directly
 
**Environment report**

```
Oh My Fish version:   6
OS type:              Linux
Fish version:         fish, version 2.7.1
Git version:          git version 2.17.1
Git core.autocrlf:    no
Your shell is ready to swim.
```

SleeplessSloth:
# Description

Call `command rm` instead of `rm` in `omf destroy` since `rm` may be aliased to something else and fail.

Fixes #717

**Environment report**

```
Oh My Fish version:   6
OS type:              Linux
Git version:          git version 2.23.0
Git core.autocrlf:    no
Your shell is ready to swim.
```

# Checklist:

### DIFF
--- a/pkg/omf/functions/bundle/omf.bundle.remove.fish
+++ b/pkg/omf/functions/bundle/omf.bundle.remove.fish
@@ -10,7 +10,7 @@ function omf.bundle.remove
       set name $argv[2]
       set bundle_contents (cat $bundle | sort -u)
 
-      rm -f $bundle
+      command rm -f $bundle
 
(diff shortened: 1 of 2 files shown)


derekstavis:
Thank you!!

PR #726, opened 2019-10-11
Update install sha256 in README.

joelburget:


scorphus:
Good catch, @joelburget. Thanks!

PR #742, opened 2020-03-09
Fixes #738 by changing the git config autocrlf test to check for what…

### LINKED ISSUE #738: Valid value "input" for git autocrlf is not handle properly in sanity check
After trying to install via the url piped directly to fish failed. I download the installer script and ran: `./install --help`. Seeing that there was a "system readiness check" I then ran:

```bash
./install --check
```
Seeing as that provided very little information I then tried:

```bash
./install --check --verbose
DEBUG: verbose turned on
DEBUG: Command 'awk' is /usr/bin/gawk
DEBUG: Command 'basename' is /usr/bin/basename
DEBUG: Command 'cp' is /bin/cp
DEBUG: Command 'cut' is /usr/bin/cut
DEBUG: Command 'date' is /bin/date
DEBUG: Command 'dirname' is /usr/bin/dirname
DEBUG: Command 'env' is /usr/bin/env
DEBUG: Command 'fish' is /usr/bin/fish
DEBUG: Command 'fold' is /usr/bin/fold
DEBUG: Command 'head' is /usr/bin/head
DEBUG: Command 'mkdir' is /bin/mkdir
DEBUG: Command 'mv' is /bin/mv
DEBUG: Command 'readlink' is /bin/readlink
DEBUG: Command 'rm' is /bin/rm
DEBUG: Command 'sed' is /bin/sed
DEBUG: Command 'sort' is /usr/bin/sort
DEBUG: Command 'tar' is /bin/tar
DEBUG: Command 'tr' is /usr/bin/tr
DEBUG: Command 'which' is /bin/which
DEBUG: Checking for a sane 'head' implementation
DEBUG: Checking for a sane 'sort' implementation
DEBUG: Checking for a working AWK interpreter
DEBUG: Verifying Git implementation is not buggy Git for Windows
DEBUG: Verifying Git autocrlf is not enabled
fatal: bad numeric config value 'input' for 'core.autocrlf' in /home/fhenderson/.gitconfig: invalid unit
```
I then checked my setting for git core.autocrlf:
```bash
git config core.autocrlf                                                                                                     209ms  [Die Feb 25 15:46:56]
input
```

Checking the Git documentation for [Formatting and Whitespace](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace), core.autocrlf has three possible values: `true`, `false` and `input`.

Examing the code in `install` 

https://github.com/oh-my-fish/oh-my-fish/blob/b2d7a444fb5f459cf4ee3354e0b52c9a116defbf/bin/install#L483-L487

we see that it is not handling the possible value `input` properly and this is what is aborting the install.

I believe that what should be tested for here is if the value of core.autocrlf is set to **false** with something like this:
```fish
 if test (cmd git config core.autocrlf) = false
    echo "Please disable core.autocrlf in your Git configuration."
  end
```

Thanks for looking into this!
Frederick

----

**Environment report**

I am trying to install a newer version after uninstalling a very old version and the install is failing, so omf is not installed.

```
$ omf doctor                                                                                                                           [Die Feb 25 15:26:28]
omf: command not found
```

    Operating System: Ubuntu 16.04.6 LTS
    Fish Version: 3.1.0
    Oh My Fish Version: not yet reinstalled
    Git version: 2.7.4

frederickjh:
# Description
Instead of trying to convert the output of `git config autocrlf` to binary, which will not always work as "input" is also  a valid value in addition to true and false, we check for the value false and exit if it is found.

Fixes # (issue)

**Environment report**

```
Oh My Fish version:   6
OS type:              Linux
Git version:          git version 2.20.1
Git core.autocrlf:    input
Your shell is ready to swim.
```

# Checklist:

I do not think that documentation changes are needed as this is a bug fix nor are code comments needed as the change is small enough to understand. Not sure how to update the SHA256 Checksum and no instructions are given in [CONTRIBUTING.md](CONTRIBUTING.md).


bin/install
```diff
@@ -481,8 +481,8 @@ function sane_environment_check
   end
 
   debug "Verifying Git autocrlf is not enabled"
-  if test (cmd git config --bool core.autocrlf; or echo false) = true
-    abort "Please disable core.autocrlf in your Git configuration."
+  if test (cmd git config core.autocrlf) = false
+    echo "Please disable core.autocrlf in your Git configuration."
```
bobthecow: changing this to `echo` means it won't exit on failure, right?

frederickjh: @bobthecow You might think so, but when git core.autocrlf was set to "input" it fail and exited the installer. Here is the output from the current installer with core.autocrlf set to input using the verbose and check options.

```fish
$  ./install --verbose --check                                                           http://ohmyfish.dev.topdroplabs.com  [Mon Mar  9 18:48:01]
DEBUG: verbose turned on
DEBUG: Command 'awk' is /usr/bin/gawk
DEBUG: Command 'basename' is /usr/bin/basename
DEBUG: Command 'cp' is /bin/cp
DEBUG: Command 'cut' is /usr/bin/cut
DEBUG: Command 'date' is /bin/date
DEBUG: Command 'dirname' is /usr/bin/dirname
DEBUG: Command 'env' is /usr/bin/env
DEBUG: Command 'fish' is /usr/bin/fish
DEBUG: Command 'fold' is /usr/bin/fold
DEBUG: Command 'head' is /usr/bin/head
DEBUG: Command 'mkdir' is /bin/mkdir
DEBUG: Command 'mv' is /bin/mv
DEBUG: Command 'readlink' is /bin/readlink
DEBUG: Command 'rm' is /bin/rm
DEBUG: Command 'sed' is /bin/sed
DEBUG: Command 'sort' is /usr/bin/sort
DEBUG: Command 'tar' is /bin/tar
DEBUG: Command 'tr' is /usr/bin/tr
DEBUG: Command 'which' is /bin/which
DEBUG: Checking for a sane 'head' implementation
DEBUG: Checking for a sane 'sort' implementation
DEBUG: Checking for a working AWK interpreter
DEBUG: Verifying Git implementation is not buggy Git for Windows
DEBUG: Verifying Git autocrlf is not enabled
fatal: bad numeric config value 'input' for 'core.autocrlf' in file /home/docker/.gitconfig: invalid unit
```
I went back and tested the original installer using the three possible values. It looks like I have this wrong. I should be testing for true not false. "false" and "input" are acceptable to the installer but "true" is not.

frederickjh: @bobthecow I when back and took another look at the original code. I believe the reason that the value input does not get processed correctly, is the `--bool` option as "input" does not seem to convert to a boolean. I revising the code to:
```fish
if test (cmd git config core.autocrlf; or echo false) = true
    abort "Please disable core.autocrlf in your Git configuration."
```
and testing with the three possible values and it does work as expected with true aborting the installer and false and input being accepted values and the installer continues.

frederickjh: @bobthecow I can revise this pull request this way if you prefer it over mine, which I find much simpler.

frederickjh:
I figured out how to generate a SHA256 checksum for the installer and updated the documentation, so once the Travis build finishes this should be ready to be reviewed.

bin/install
bobthecow: yeah, i like that too, thanks.

frederickjh: Sorry, @bobthecow  I am not sure I understand what you mean. Do you like the fix I coded or the idea to remove the `--bool` option?

bobthecow: Yes, I mean let's update it to use the code in https://github.com/oh-my-fish/oh-my-fish/pull/742/files#r389878048 

Thanks!

frederickjh: Done.

PR #767, opened 2020-06-30
Get when a file was deleted.

damani42:
When working on a large project and maintaining several branches, it is
intersting to be able to know from which branch a file was deleted.

### DIFF
--- a/lib/git/git_when_file_was_deleted.fish
+++ b/lib/git/git_when_file_was_deleted.fish
@@ -1,0 +1,27 @@
+function git_when_file_was_deleted -d "Get the sha and branch when a file was deleted"
+    if not set -q argv[1]
+        return 1
+    end
+
+    if not git_is_repo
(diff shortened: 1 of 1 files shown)


damani42:
I would like to know if it's possible to have a review please?

bobthecow:
This isn't really the sort of thing `lib/git` is meant for. These are core libraries used internally by OMF for managing packages.

The ideal place to put things like this is in a package (which can then be added to OMF, and managed by these core libraries 😛).

faho:
Another that was basically already rejected. I'd close it and add it to a git-util plugin or something?

scorphus:
Dear @damani42,

Thank you so much for taking the time to contribute.

I have a suggestion. Your solution is useful not only for OMF and Fish users but Git users in general. So, I suggest you publish it as a git subcommand along with others under such topic: https://github.com/topics/git-subcommand. That way, many more users would benefit from it 🙂

PR #768, opened 2020-07-01
fix broken link in Themes.md

zyctree:
fix the broken link

it is caused by https://github.com/oh-my-fish/oh-my-fish/pull/694

here is a simple check
* <https://github.com/oh-my-fish/oh-my-fish/blob/master/docs/Themes.md#pure>
* <https://github.com/oh-my-fish/oh-my-fish/blob/master/docs/Themes.md#pure----->


bobthecow:
Thank you for the PR. This was fixed by regenerating the file in #806

PR #775, opened 2020-08-16
Fix double slash in index path

### LINKED ISSUE #774: Index update immediately removed due to double slash
Due to the extra trailing slash in variable like `XDG_CACHE_HOME`, the repository index folder is immediately deleted.

## Troubleshoot
This is confirmed with the following modification,

```diff
diff --git a/pkg/omf/functions/index/omf.index.update.fish b/pkg/omf/functions/index/omf.index.update.fish
index 1e2c69f..3539b51 100644
--- a/pkg/omf/functions/index/omf.index.update.fish
+++ b/pkg/omf/functions/index/omf.index.update.fish
@@ -51,8 +51,11 @@ function omf.index.update -d 'Update package indexes'
     set -e repositories[1..3]
   end
 
+echo "Valid paths: $valid_paths";
+
   # Remove repositories not in the lists.
   for path in (omf.index.path)/*
+echo ".. evaluating path: $path";
     if not contains -- $path $valid_paths
       command rm -rf $path
     end
```

This returning 

```
$ omf search "bobthefish"
Updating https://github.com/oh-my-fish/packages-main master... Done!
Valid paths: /home/user/.cache//omf/https___github.com_oh-my-fish_packages-main_master
.. evaluating path: /home/user/.cache/omf/https___github.com_oh-my-fish_packages-main_master
Plugins

Themes
```

Notice the additional slash.

## Possible Symptoms
1. `omf search` returns empty.
2. `omf update` not creating any file on cache dir (`omf.index.path`).
3. First time installation cause "Repository 'default' do not exist", maybe related to [trancef0rmer's comment](https://github.com/oh-my-fish/oh-my-fish/issues/454#issuecomment-461761466).

**Environment report**

```
Oh My Fish version:   7
OS type:              Linux
Git version:          git version 2.25.4
Git core.autocrlf:    no
Your shell is ready to swim.
```

Abdillah:
# Description
`XDG_*` or `HOME` variables might contains extra slash at the end of path (I don't know if this is against specification). This double slash might disturb comparison and make them false positive. The consequence is as described in the issue.

The error can happen at the first time OMF installed thus might block potential new user.

Fixes #774 

**Environment report**

```
Oh My Fish version:   7
OS type:              Linux
Git version:          git version 2.25.4
Git core.autocrlf:    no
Your shell is ready to swim.

```

# Checklist:

- [ ] New and existing tests pass locally with my changes 
- [ ] I have updated the SHA256 checksum for the install script

### DIFF
--- a/pkg/omf/functions/index/omf.index.path.fish
+++ b/pkg/omf/functions/index/omf.index.path.fish
@@ -1,5 +1,5 @@
 function omf.index.path -d 'Get the path to the local package index'
   set -q XDG_CACHE_HOME
-    and echo "$XDG_CACHE_HOME/omf"
-    or echo "$HOME/.cache/omf"
+    and echo (string trim -r -c "/" $XDG_CACHE_HOME)"/omf"
+    or echo (string trim -r -c "/" $HOME)"/.cache/omf"
 end
--- a/pkg/omf/functions/index/omf.index.path.fish
+++ b/pkg/omf/functions/index/omf.index.path.fish
@@ -1,5 +1,5 @@
 function omf.index.path -d 'Get the path to the local package index'
   set -q XDG_CACHE_HOME
-    and echo (string trim -r -c "/" $XDG_CACHE_HOME)"/omf"
-    or echo (string trim -r -c "/" $HOME)"/.cache/omf"
+    and echo (echo $XDG_CACHE_HOME | sed 's:/*$::')"/omf"
+    or echo (echo $HOME | sed 's:/*$::')"/.cache/omf"
 end


faho:
Either this or changing to using `realpath` before comparing paths should be merged.

This is a simple fix, *however* because of `string` it requires fish 2.3.0 or later, while the README currently mentions compatibility with 2.2.0.

@scorphus Is fish 2.2.0 compatibility still wanted? If yes, this would have to use `sed` or something instead.

scorphus:
I think 2.2 compatibility is desired. Ubuntu 16.04 will remain in LTS until 2024, IIRC.

faho:
>I think 2.2 compatibility is desired. Ubuntu 16.04 will remain in LTS until 2024, IIRC.

There's an [official PPA](https://launchpad.net/~fish-shell/+archive/ubuntu/release-3) for Ubuntu. Personally, I'd just direct users to that.

But, well, it's a maintainer decision. In that case this PR needs to be rewritten to use `sed` or you'll have to use `realpath` (and not the builtin version).

scorphus:
Dropping 2.2.0 support is probably another, separate discussion. So far I guess we keep supporting it.

Wouldn't `sed`'s different behaviour across systems be a problem? How about something like this:

```fish
function omf.path.join -a prefix -a suffix
  echo $prefix | grep -qE "/\$"
    and echo $prefix$suffix
    or echo $prefix"/"$suffix
end
```

It's not as fast as `string trim` but it's a bit faster than `sed`.

Abdillah:
> you'll have to use realpath (and not the builtin version).

Why not built-in fish `realpath`?

faho:
>Why not built-in fish realpath?

Well, because it's not in 2.2.0.

And for compatibility reasons it was decided that any realpath *command* takes precedence (instead of picking a different name for the builtin, which we probably should have just done), so if you do `realpath something` you'll get the command (that might not even exist!), and `builtin realpath something` isn't supported in fish 2.2.0 (only since 2.4.0).

faho:
>Wouldn't sed's different behaviour across systems be a problem?

@scorphus In this case, i.e. simply trimming a "/"? No.

```fish
echo thing | sed 's:/*$::'
```

should "just work" everywhere?

pkg/omf/functions/index/omf.index.path.fish
```diff
@@ -1,5 +1,5 @@
 function omf.index.path -d 'Get the path to the local package index'
   set -q XDG_CACHE_HOME
-    and echo "$XDG_CACHE_HOME/omf"
-    or echo "$HOME/.cache/omf"
+    and echo (string trim -r -c "/" $XDG_CACHE_HOME)"/omf"
+    or echo (string trim -r -c "/" $HOME)"/.cache/omf"
```
scorphus, line 4:
```suggestion
    and echo (echo $XDG_CACHE_HOME | sed 's:/*$::')"/omf"
    or echo (echo $HOME | sed 's:/*$::')"/.cache/omf"
```

scorphus (changes requested):
@Abdillah, please consider the requested changes.

Thanks, @faho.

pkg/omf/functions/index/omf.index.path.fish
Abdillah: Sure.

scorphus (approved):
Thanks, @Abdillah!

PR #797, opened 2020-12-06
Move build from TravisCI to GitHub Actions

scorphus:
Please review, comment, and suggest improvements.


.github/workflows/build.yml
```diff
@@ -0,0 +1,67 @@
+name: Build
+on: [push, pull_request]
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        ubuntu-release:
```
bobthecow, line 8: Is it worth including macOS in the matrix like we had on Travis? Maybe save it for another PR?

bobthecow (approved):
Thanks for taking this on, and thanks for building a proper test matrix this time :) 

I'm a bit worried about the results on Fish 2 (and "stock" for most Ubuntus). The test logs look a _lot_ like failures, but the builds aren't failing. Looking at the Linux job on Travis, it was doing this too. Do we need to do something to fix that?

Given that it was already doing this on Travis, I don't think it's a blocker, but we should figure out why it's spammy and looks like errors.

.github/workflows/build.yml
scorphus: Yes, definitely! Maybe the same PR. Sorry I forgot to comment on that. Will do below.

scorphus:
> Thanks for taking this on, and thanks for building a proper test matrix this time :)

❤️ You're welcome! I took the chance to improve and fill the gaps I could see.

> I'm a bit worried about the results on Fish 2 (and "stock" for most Ubuntus). The test logs look a lot like failures, but the builds aren't failing. Looking at the Linux job on Travis, it was doing this too. Do we need to do something to fix that?

Good point, I'm not sure yet. I noticed that too but didn't dig into none of them. I spent quite some time with the Slack integration, searching for one action that could do the job without bringing in any dependency (hence the ugliness and the hardcoded, fixed ref in there). I almost gave up and even thought about writing a shell script. Boah... 😅 

> Given that it was already doing this on Travis, I don't think it's a blocker, but we should figure out why it's spammy and looks like errors.

I'll definitely spend more time on this PR, making sure there're no false positive builds, and including MacOS in that matrix — which I knowingly removed from `tools/ci-install-fish.sh` 🙃

🙌 Thanks for your inputs! Please LMK if you see anything else that could be done better/easier.

scorphus:
@bobthecow: I added a few more commits, fixing some issues with older Fish versions. There's also one commit for test improvements. Please LMK if I should split them into separate PRs. Please kindly review this again when you have the time. And for that I thank you in advance.

bobthecow:
Thanks @scorphus!

PR #806, opened 2021-01-22
Make `generate-themes-doc` usable and update Themes.md

scorphus:
I didn't have much time to invest in `tools/generate-themes.doc.fish`, unfortuntely. It's far from perfect, but fairly usable. Please, refer to instructions in the header.

Fix #795, #800, #771

PR #809, opened 2021-01-24
Themes.md: Show a table of themes instead of a long list

scorphus:
Ref #806

### DIFF
--- a/docs/Themes.md
+++ b/docs/Themes.md
@@ -1,100 +1,142 @@
 # Available themes
-- [agnoster](#agnoster)
-- [aight](#aight)
-- [ays](#ays)
-- [batman](#batman)
-- [beloglazov](#beloglazov)
-- [bira](#bira)
-- [bobthefish](#bobthefish)
-- [bongnoster](#bongnoster)
-- [boxfish](#boxfish)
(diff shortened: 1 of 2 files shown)


scorphus:
@bobthecow: Sorry for the extra noise, I guess I created the PR too soon.

I experimented with the number of columns and think 7 is okay, but that's MO. I'd love to read your thoughts.

docs/Themes.md
```diff
@@ -1,100 +1,142 @@
...
-- [will](#will)
-- [wolf-theme](#wolf-theme)
-- [yimmy](#yimmy)
-- [zeit](#zeit)
-- [zephyr](#zephyr)
-- [zish](#zish)
-
-# agnoster
+
+<!-- ATTENTION: This file is auto-generated! Changes will be lost! -->
```
bobthecow, line 3: 🎉

bobthecow (approved):
:shipit:

bobthecow:
> I experimented with the number of columns and think 7 is okay, but that's MO. I'd love to read your thoughts.

Yeah, that looks good to me. We can always adjust later if we need to. Thanks!

PR #810, opened 2021-01-24
workflow: Send no notification on Slack

scorphus:
This is somehow useless as secrets are not passed to action runners when the workflow is triggered from a forked repository (for obvious reasons)


bobthecow (approved):
😢

PR #823, opened 2021-03-02
Improve nonstandard uninstall events for backwards compatability

fdeitylink:
# Description

This changes the `remove` script to check if an `uninstall.fish` script exists before running it. It also fires `{$pkg}_uninstall` and `uninstall_$pkg` events whether or not the file exists. This improves support for nonstandard uninstall hooks.

**Environment report**

```
Oh My Fish version:   7-31-ga579039
OS type:              Linux
Fish version:         fish, version 3.1.2
Git version:          git version 2.25.1
Git core.autocrlf:    no
Your shell is ready to swim.
```

# Checklist:

### DIFF
--- a/docs/en-US/Packages.md
+++ b/docs/en-US/Packages.md
@@ -120,6 +120,7 @@ Inside this hook, you can access two package-related variables:
 Packages can use this hook to clean up custom resources, etc.
 
 > Note: for backwards-compatibility, uninstall hooks will also be run if they are located at `uninstall.fish` in the package root.
+> Hooks may also be triggered by listening for the event `{$package}_uninstall` or `uninstall_$package`.
 
 # Make it public
 
--- a/pkg/omf/functions/packages/omf.packages.remove.fish
+++ b/pkg/omf/functions/packages/omf.packages.remove.fish
@@ -17,8 +17,11 @@ function omf.packages.remove -a pkg
 
     # Run uninstall hook first.
     omf.packages.run_hook $path uninstall
-    source $path/uninstall.fish 2> /dev/null;
-      and emit uninstall_$pkg
+    if test -f $path/uninstall.fish
+      source $path/uninstall.fish 2> /dev/null
+    end
+    emit uninstall_$pkg
+    emit {$pkg}_uninstall
 
     if command rm -rf $path
       omf.bundle.remove "package" $pkg


scorphus (approved):
🤘

bobthecow:
Why both uninstall_* and *_uninstall? Are there examples in the wild with both?

fdeitylink:
Yes, not all packages are consistent with the hook name

scorphus:
The inconsistencies — or incompatibilities — are such as the one in [fzf.fish](https://github.com/PatrickF1/fzf.fish/blob/f94544c67e57b9e8cb2cd3cf4c07c5039e77e16b/conf.d/fzf.fish#L39-L51):

```fish
function _fzf_uninstall --on-event fzf_uninstall
  # ...
end
```

And the one in soon-to-be changed [proxy](https://github.com/oh-my-fish/plugin-proxy):

```fish
function _proxy_uninstall --on-event uninstall_proxy
  # ...
end
```

We prolly want `proxy` to be supported by other frameworks, so that might actually be changed to:

```fish
function _proxy_uninstall --on-event proxy_uninstall
  # ...
end
```

We're also prolly not aguing as to which one is the default or why has `*_uninstall` been brought up in the first place.

**Important**: once this gets merged, we need to “fix” [bang-bang](https://github.com/oh-my-fish/plugin-bang-bang) as it has both an event-triggered uninstall function and an uninstall hook.

bobthecow:
Is it better to settle on one (probably whatever fisher uses), then update the plugins that don’t match?

scorphus:
If we're sure all plugins living out there adhere to that standard, I don't see why not. That's something hard to do, so, we might as well make an announcement in README so that plugin authors can update their code and users have a chance to open more detailed, to-the-point issues.

scorphus:
Or, we can deprecate `ununstall_*`, include that in the docs and in the changelog, display an alert whenever a user installs such a plugin and remove it in a subsequent release.

bobthecow:
They seems like a reasonable approach.

faho:
This was approved by one reviewer, so it seems like it can be merged?

scorphus:
I need to give this one another test and review round. Thanks for the reminder, @faho

scorphus:
> Or, we can deprecate `uninstall_*`, include that in the docs and in the changelog, display an alert whenever a user installs such a plugin and remove it in a subsequent release. That would also give us time to go over all the plugins.

I'll create an issue to deprecate `uninstall_*` after merging this PR.

scorphus:
Thank you very much for your contribution, @fdeitylink. It's a shame it took this long to get merged. I really hope OMF gets more contribution from you!

PR #824, opened 2021-03-03
Installation with user/repo syntax

### LINKED ISSUE #818: Proposal: Assume github.com when no package is found
Change the package installation in a way that, when a user issues `omf install user/repo`, assume the user intends to install [https://github.com/user/repo]().

fdeitylink:
# Description

Allows installation from GitHub repositories with `user/repo` syntax. Shorthand for URL syntax, more or less, with GitHub as the forge.

Fixes #818 

**Environment report**

```
Oh My Fish version:   7-35-g332ca07
OS type:              Linux
Fish version:         fish, version 3.1.2
Git version:          git version 2.25.1
Git core.autocrlf:    no
```

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,11 +86,12 @@ Update Oh My Fish, all package repositories, and all installed packages.
 - For selective package update, list only the names of packages you wish to
   update. You may still include "omf" in the list to update the core as well.
 
-#### `omf install` _`[<name>|<url>]`_
+#### `omf install` _`[<name>|<url>|<user/repo>]`_
 
 Install one _or more_ packages.
 
 - You can install packages directly by URL via `omf install URL`
+- You can install packages from a GitHub repository via `omf install user/repo`
 - When called without arguments, install missing packages from [bundle](#dotfiles).
 
 #### `omf repositories` _`[list|add|remove]`_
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -107,7 +108,7 @@ Apply a theme. To list available themes, type `omf theme`. You can also [preview
 
 #### `omf remove` _`<name>`_
 
-Remove a theme or package.
+Remove a theme or package. If a package was installed via `user/repo`, use `repo` for `name`.
```
bobthecow, line 111: Should we refer to packages installed this way as user/repo? Otherwise we have potential collisions as multiple users might have the same repo name, and repo names might collide with package names from the registry. 

Thinking about it, maybe a good approach would be to install these as `@user/repo` in the filesystem, while still allowing them to be referred to as either `user/repo` or `repo` as long as there is no ambiguity.

fdeitylink: That's a good point. I have to test if installing to `$OMF_PATH/pkg/user/repo` would break anything since packages are expected to be found in `$OMF_PATH/pkg`. Additionally, if we're going with that, I think we shouldn't use `omf.packages.name` to set a package name when installing via URL or `user/repo` syntax. In my opinion it creates a potentially confusing disparity between installation and uninstallation. Thoughts?

bobthecow: Oh, yeah. I was thinking that `@user/repo` would keep the ones installed via this mechanism from colliding, but I guess we have a similar problem with packages installed by URL.

fdeitylink: I'm thinking the installation subdirectory could map basically 1 to 1 with the installation "name" e.g. anything installed from a URL is installed to `$OMF_PATH/pkg/url`. It's not wholly uncommon, prevents conflicts, and usually users shouldn't need to muck around in there.

scorphus: The separation of plugins in namespaces to avoid collisions seems like a good idea, but I think it shouldn't prevent this PR to be merged, as the functionality is already there — and so is the collision. I'm merging this one and opening an issue for when we'd like to improve pkg installation and mitigate collisions.

scorphus:
Many thanks, @fdeitylink! 🙌

PR #844, opened 2021-07-11
Remove archived and unsupported shellder

JAicewizard:
# Description

Simple removal of shellder, in the OMF list, since it is no longer supported and doesnt work properly anymore.

This theme lists the binaries and giz, svn and gh in the prompt rendering it in the "you really shouldn't use this anymore" category.


faho:
As far as I know, THEMES.md is autogenerated from https://github.com/oh-my-fish/packages-main/tree/master/packages, and this should therefore be removed there.

(and a "check if archived or deleted" script would probably be a useful thing)

faho:
This was done in https://github.com/oh-my-fish/packages-main/pull/151, THEMES.md just needs to be regenerated.

scorphus:
Thanks for contributing, @JAicewizard. oh-my-fish/packages-main#151 included not only `shellder` but a few other pkgs in the same situation.

JAicewizard:
Yup thats a much better contribution. thanks @faho for the better fix.

PR #851, opened 2021-07-31
doctor: check if Git remotes match db URLs

scorphus:
Continuing @oranja's work.

### DIFF
--- a/pkg/omf/functions/repo/omf.repo.remote_uri.fish
+++ b/pkg/omf/functions/repo/omf.repo.remote_uri.fish
@@ -1,0 +1,13 @@
+function omf.repo.remote_uri -a repo_dir -a branch
+  test -z "$repo_dir"; and set repo_dir "$PWD"
+  set -l git_dir (command git -C "$repo_dir" rev-parse --show-toplevel 2>/dev/null)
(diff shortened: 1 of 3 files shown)


pkg/omf/functions/repo/omf.repo.remote_uri.fish
```diff
@@ -0,0 +1,13 @@
+function omf.repo.remote_uri -a repo_dir -a branch
+  test -z "$repo_dir"; and set repo_dir "$PWD"
+  set -l git_dir (command git -C "$repo_dir" rev-parse --show-toplevel 2>/dev/null)
+    and test "$git_dir" = "$repo_dir"
+    or return 1
```
scorphus, line 5: @oranja, these are the only two of lines I've got a comment for. I tested this change in many different ways, one of them using a soft link to a plugin. In this case, nothing is returned and I end up with something like:

    Warning: The remote URI for apt doesn't match the URI in OMF's database.
    Remote URI:
    Index URI: https://github.com/oh-my-fish/plugin-apt

A missing remote URI, even though the plugin works fine.

Perhaps there could be a previous check for such cases with an appropriate warning.

What do you think?

PR #859, opened 2021-09-28
docs(lib): rm unnecessary img tag from readme

jjangga0214:


jjangga0214:
I see a little bit of gap by empty img tag. Does it really needed and cannot be done by other mean?

derekstavis (approved):

PR #860, opened 2021-09-28
ci(build): ignore if only **.md is editted

jjangga0214:
REF: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-paths


jjangga0214:
I believe this is a simple pr, so it would not take much time to review. 
ping?

derekstavis (approved):

PR #863, opened 2021-10-19
Require bundle package after install

Detailer:
# Description

After a package dependency was installed from the bundle file, the dependency package was not available for use by the plugin to be installed,. Thus, causing issues with plugins which relied on dependency packages during installation. Added require to packages installed from bundle file as dependencies. 

Fixes Related Issue: https://github.com/oh-my-fish/plugin-weather/issues/38

**Environment report**

```
Oh My Fish version:   7-35-g3c0c810
OS type:              Linux
Fish version:         fish, version 3.3.1
Git version:          git version 2.32.0
Git core.autocrlf:    no
Your shell is ready to swim.
```

# Checklist:

- [*] My code follows the [style guidelines](https://github.com/oh-my-fish/oh-my-fish/blob/master/CONTRIBUTING.md#code-style) of this project
- [*] I have performed a self-review of my own code
- [*] I have commented my code, particularly in hard-to-understand areas
- [*] I have made corresponding changes to the documentation
- [*] New and existing tests pass locally with my changes

### DIFF
--- a/pkg/omf/functions/bundle/omf.bundle.install.fish
+++ b/pkg/omf/functions/bundle/omf.bundle.install.fish
@@ -22,8 +22,11 @@ function omf.bundle.install
         omf.packages.install $name_or_url;
           and begin
             test $type = package
-              and require $name
-              or echo "Failed to require package: $name"
+              and begin
+                require $name
+                  or echo "Failed to Require $package"
+              end
+              or true
           end
           or set error
       end
--- a/pkg/omf/functions/bundle/omf.bundle.install.fish
+++ b/pkg/omf/functions/bundle/omf.bundle.install.fish
@@ -20,6 +20,7 @@ function omf.bundle.install
 
       if not contains $name $packages
         omf.packages.install $name_or_url;
+          and require $name_or_url
           or set error
       end
     end
--- a/pkg/omf/functions/bundle/omf.bundle.install.fish
+++ b/pkg/omf/functions/bundle/omf.bundle.install.fish
@@ -20,8 +20,11 @@ function omf.bundle.install
 
       if not contains $name $packages
         omf.packages.install $name_or_url;
-          and require $name_or_url
-          or set error
+        and begin 
+          test $type = package
+            and require $name
+        end
+        or set error
       end
     end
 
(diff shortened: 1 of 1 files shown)


pkg/omf/functions/bundle/omf.bundle.install.fish
```diff
@@ -17,9 +17,18 @@ function omf.bundle.install
       test -n "$name_or_url"; or continue
 
       set name (omf.packages.name $name_or_url)
-
+
```
scorphus, line 20:
```suggestion
```

pkg/omf/functions/bundle/omf.bundle.install.fish
```diff
@@ -17,9 +17,18 @@ function omf.bundle.install
...
         omf.packages.install $name_or_url;
+        
+          # require if not a theme
+          and begin 
+            test $type = package; and begin
+              require $name
+                or echo "Failed to require $name"
+            end
+            or echo "Skipped require $name"
+          end
```
scorphus, line 31: As no messages need to be displayed, I guess those echo lines could be omitted. What do you think?

```suggestion
          and begin
            test $type = package;
              and require $name;
              or true
          end
```

It would be useful informing of failures while requiring packages, but that's prolly a subject for another PR.

scorphus (approved):
We're getting there. Please consider my observations and comment on them.

pkg/omf/functions/bundle/omf.bundle.install.fish
Detailer: Yes, I agree the second suggest code block is better. But I think we should inform if any failure while requiring occurred. Maybe a if block would be more readable?

scorphus: True. We're not just solving oh-my-fish/plugin-weather#38. Agreed 👍

scorphus:
As I updated the build environments, if you rebase on top of current master, the build should stop failing.

Detailer:
@scorphus Made code block more readable, added error message in case `require $pacakge` fails.

Detailer:
My bad, the error flow was not proper in the last commit. Fixed now.

derekstavis (approved):

scorphus (approved):

scorphus:
Thanks for your contribution, @Detailer! Sorry for the delay.

Thanks for the nudge, @derekstavis.

PR #892, opened 2022-03-26
omf.cli.install: Set theme when installing from URL

scorphus:
Set theme when installed from URL (or user/repo)

### DIFF
--- a/pkg/omf/functions/cli/omf.cli.install.fish
+++ b/pkg/omf/functions/cli/omf.cli.install.fish
@@ -14,8 +14,11 @@ function omf.cli.install
         and require $package
 
       # If package is a theme, set it to active.
-      if contains -- $package (omf.packages.list --theme)
+      set -l themes (omf.packages.list --theme)
+      if contains -- $package $themes
         omf.theme.set $package
+      else if set -l ind (contains -i -- (omf.packages.name $package) $themes)
+        omf.theme.set $themes[$ind]
       end
 
       test $status != 0;


scorphus:
Hey, @faho! Would you kindly drop some of your words of wisdom? Thanks in advance!

pkg/omf/functions/cli/omf.cli.install.fish
```diff
@@ -14,8 +14,11 @@ function omf.cli.install
         and require $package
 
       # If package is a theme, set it to active.
-      if contains -- $package (omf.packages.list --theme)
+      set -l themes (omf.packages.list --theme)
+      if contains -- $package $themes
         omf.theme.set $package
+      else if contains -- (omf.packages.name $package) $themes
+        omf.theme.set (omf.packages.name $package)
```
faho, line 21: This of course calls `omf.packages.name` twice. Which calls `basename` and, because y'all still support fish 2.2.0, `sed`.

There are two alternatives:

1. Store the package name in a variable
2. Use `contains -i`:

```fish
else if set -l ind (contains -i -- (omf.packages.name $package) $themes)
    omf.theme.set $themes[$ind]
```

faho: (I make no guarantee that this works with fish 2.2.0, it's been years since I have used that)

faho (commented):
I can really only give a minor comment. I don't know the meaning or context here, I can just tell you that you could save a function call, which could save a millisecond or so. It's probably not important, unless this happens in a hot path?

pkg/omf/functions/cli/omf.cli.install.fish
scorphus: Works nicely with 2.2.0! Many thanks, @faho 🙌

PR #894, opened 2022-04-10
Update init.sh

BETAIDK:
add a new command for fish
this command called config what he does is without typing cd ~/.config/fish, just move you to this directory


scorphus:
Functions are mostly a very personal thing. Many functions fall into such category and the one suggested in this PR is an example.

OMF tries to include the least amount of functions as possible. In view of that, it would be advisable to keep functions like this inside `~/.config/fish/functions`.

PR #904, opened 2022-06-16
Fix cutting of paths with equals signs

pinselimo:
# Description

In the install script paths containing equals signs (e.g. as argument to the ```--offline``` flag) are truncated. ```cut``` is set to split by equals sign and take the second field. This ignores path contents after a second equals sign. Currently this prevents the installation on NixOS as the nix store path automatically contains a "=" and is used as offline source.

This PR instead selects all but the first field and equals sign, leaving the passed paths as intended.

Note: It also fixes similar errors caused by path arguments containing an equals sign.

**Environment report**

```
Oh My Fish version:   7
OS type:              Linux
Fish version:         fish, version 3.4.1
Git version:          git version 2.36.0
Git core.autocrlf:    no
Your shell is ready to swim.
```

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas


bin/install
```diff
@@ -54,13 +54,13 @@ Options:
         set -g OMF_CHANNEL dev
 
       case '--channel=*'
-        abort "Unknown release channel \""(echo "$argv[1]" | command cut -d= -f2)"\"."
+        abort "Unknown release channel \""(echo "$argv[1]" | command cut -d= -f1 --complement)"\"."
```
scorphus, line 57: Can you please show a working example of `cut -d= -f1 --complement`?

I tried with `echo --output=foo=bar | command cut -d= -f1 --complement` which gives the following error:

```
cut: illegal option -- -
usage: cut -b list [-n] [file ...]
       cut -c list [file ...]
       cut -f list [-s] [-d delim] [file ...]
```

pinselimo: Apparently the ```--complement``` options is [only available in GNU cut](https://www.gnu.org/software/coreutils/manual/html_node/cut-invocation.html) and not POSIX. My bad, I am surry, I used the help page to find a solution.

The POSIX compatible way is:

~~~
$ echo --output=foo=bar | command cut -d= -f2-
foo=bar
~~~

Thanks for catching this!

scorphus (approved):
Thanks, @pinselimo 👍

PR #925, opened 2023-01-14
Fix #924: Checksum isn't up to date in readme

### LINKED ISSUE #924: Installer checksum in README.md isn't up to date.

jeanlucthumm:
Seems like this is likely to happen again in the future -- people will remember to update the SHA256 file but not the`README.md`. Couldn't find a way to embed a file into markdown, maybe add this as another checkbox in the PR template? Or remove the hash from the README.md entirely?

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ fish install --path=~/.local/share/omf --config=~/.config/omf
 You can verify the integrity of the downloaded installer by verifying the script against [this checksum](bin/install.sha256):
 
 ```
-429a76e5b5e692c921aa03456a41258b614374426f959535167222a28b676201 install
+curl -sL https://raw.githubusercontent.com/oh-my-fish/oh-my-fish/master/bin/install.sha256 | shasum -a 256 --check
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,7 +44,7 @@ fish install --path=~/.local/share/omf --config=~/.config/omf
 You can verify the integrity of the downloaded installer by verifying the script against [this checksum](bin/install.sha256):
 
 ```
-bb1f4025934600ea6feef2ec11660e17e2b6449c5a23c033860aed712ad328c9 install
+429a76e5b5e692c921aa03456a41258b614374426f959535167222a28b676201 install
```
scorphus, line 47: This was missed since the last changes in the `install` script and will be missed again. So better to only include the SHA in one place. There said, I'd rather have it like the following:

```suggestion
curl -sL https://raw.githubusercontent.com/oh-my-fish/oh-my-fish/master/bin/install.sha256 | shasum -a 256 --check
```

scorphus (approved):
Hey, good catch. Thanks!

Can you please consider my suggestion below.

scorphus (approved):

PR #927, opened 2023-02-13
require: fix conf.d support

derekstavis:
# Description

Some time ago support for conf.d was added but it is currently broken if the package does not have an `init.fish` file, given that `require` enumerates packages through a glob on the existence of `init.fish`.

This fixes this requirement by globbing both `init.fish` and `conf.d` entries in the `OMF_PATH`. This does generate duplicated entries, but given that `require` already protects against double requires, this should be safe.


A1exKH (approved):
LGTM.

PR #954, opened 2025-01-27
Fix #940: "Installer does not check `$OMF_PATH` permissions"

### LINKED ISSUE #940: Installer does not check `$OMF_PATH` permissions
For the package installation to work, `$OMF_PATH/pkg` needs to be writable. Both `bin/install` and the `omf install` command will fail when that directory is not writable by the user. Installation through Git will not lead to this situation, but some instances of offline install (where the OMF source is copied to `$OMF_PATH`) can lead to a non-writable `pkg` directory. I have encountered this issue while installing OMF as a NixOS package. The Nix package manager always stores downloaded files as read-only files. Other users have encountered this as well, see https://github.com/NixOS/nixpkgs/issues/212158

My suggestion for a solution for the problem would be to run `chmod -R u+w $OMF_PATH` in the `install_offline` function after copying from `$OFFLINE_PATH` to `$OMF_PATH`.

**Environment report**

Run the command `omf doctor` and paste the output here. Example:

```
OS type:              Linux
Fish version:         fish, version 3.6.1
Git version:          git version 2.40.1
Git core.autocrlf:    no
Your shell is ready to swim.
```

yusi1:
# Description
This fixes offline installation of oh-my-fish on NixOS (#940) where the Nix package manager stores $OMF_PATH as read-only. This makes $OMF_PATH writable again after copying.

Fixes #940.

**Environment report**
```
Oh My Fish version:   7
OS type:              Linux
Fish version:         fish, version 3.7.1
Git version:          git version 2.48.1
Git core.autocrlf:    no
Your shell is ready to swim.
```

# Checklist:

- [X] My code follows the [style guidelines](https://github.com/oh-my-fish/oh-my-fish/blob/master/CONTRIBUTING.md#code-style) of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] New and existing tests pass locally with my changes 
- [X] I have updated the SHA256 checksum for the install script


derekstavis:
Is this common practice in NixOS? I'm not very familiar with it, but is there a particular reason why it stores as read only?

yusi1:
> Is this common practice in NixOS to chmod directories? I'm not very familiar with it, but is there a particular reason why it stores as read only?

I am unsure. I do not use NixOS.

kfkonrad:
Closing this PR because there's fix in the downstream nixpkgs repo (https://github.com/NixOS/nixpkgs/pull/269493)

PR #955, opened 2025-02-23
reimplement fish-spec for nicer output

kfkonrad:
# Description

provide a new implementation for fish-spec with more verbose and colored output and more assertion functions

**Environment report**

```
Oh My Fish version:   7-61-g12fd9cb
OS type:              Darwin
Fish version:         fish, version 3.7.1
Git version:          git version 2.48.1
Git core.autocrlf:    no
Your shell is ready to swim.
```

# Checklist:


pkg/fish-spec/framework/color.fish
```diff
@@ -0,0 +1,35 @@
...
+  if test $total -eq $failed -a $total -eq 0
+    echo::success $argv[3..]
+  else if test $total -eq $failed
+    echo::failure $argv[3..]
+  else if test $failed -gt 0
+    echo::mixed $argv[3..]
+  else
+    echo::success $argv[3..]
+  end
+end
```
derekstavis, line 35: we should prolly namespace these functions and prepend `__` to avoid clashes with user defined code. e.g.: 

```fish
__fish_spec_echo_autocolor
```

pkg/fish-spec/functions/fish-spec.fish
```diff
@@ -1,67 +1,95 @@
...
+  functions -e (functions | string match -r '^(before_all|after_all|before_each|after_each|it_.*)$')
 end
 
-function __fish-spec.current_time
-  if test (uname) = 'Darwin'
-    set filename 'epoch.osx'
-  else
-    set filename 'epoch.linux'
+function __fish_spec_run_test_function -a test_func
+  set test_func_human_readable (string replace 'it_' 'IT ' $test_func | string replace -a '_' ' ')
```
derekstavis, line 75: ngl, but the uppercase `IT` throws me off every time as it reads like "IT department" 😂

derekstavis: something I was thinking we could do instead is to not send a LF at the end of the line and instead show the line with a "loading state" then rewrite the line to have a checkmark in front, e.g."

```
echo -n "⏳ it does something cool"
sleep 5
echo -n "\r"
echo -n "✅ it does something cool"
```

kfkonrad: I've implemented this, let me know if works as you'd expect

derekstavis:
I think builds are not failing when a test fails:

image

kfkonrad:
I did that on purpose. Weird as it sounds with this test case I want to test that an assertion can fail successfully. My hack (which the log output in lines 65 and 69 hint at) is decrementing the failed assertion counter to ensure the overall suite still passes

There's also an assertion right after the decrement operation to check whether the counter has the expected value - to avoid surprises from the assertion counter being manipulated in a way the tests don't expect

pkg/fish-spec/framework/assertions.fish
```diff
@@ -0,0 +1,189 @@
...
+    'Assertion failed: File \"$first\" contains a string matching the pattern \"$second\".' \
+    'not grep -qE "$second" $first'
+end
+
+function __fish_spec_in_array_helper -a value
+  for item in $__fish_spec_assertion_array
+    if test "$item" = "$value"
+      return 0
+    end
+  end
```
derekstavis, line 163: is there a particular reason to not use `contains` instead here?

derekstavis (approved):
This looks great!! Output looks much better and the modern code makes it much easier to understand. Left one non-blocking comment! Thanks for working on this one!

pkg/fish-spec/framework/assertions.fish
kfkonrad: I didn't know `contains` existed :)
I've reworked my code to use `contains` instead. It's shorter and probably faster for large arrays

PR #969, opened 2026-08-27
fix(omf): critical CLI bugs on fish 3/4

mhenrixon:
# Description

Stacked on #966 (includes it). The CI-matrix PR #968 now sits on top of this one, because extending the matrix to fish ≥ 4.1 (Alpine images: no `git`, no `column`, tarball checkout) needs these fixes first. Each fix has a spec under `pkg/omf/spec/` that failed before the change (196 assertions total, green on fish 4.8.1).

- `omf.packages.new`: `exit` killed the interactive shell when OMF paths were missing; unconditional `popd` popped the caller's dir stack.
- `omf.packages.list`: `set -l builtin_packages` at file scope was empty inside the function, so `omf`/`fish-spec` were listed and `omf update` tried to update them.
- `omf.packages.valid_name`: rejected names not starting with a lowercase ASCII letter (`omf remove Bobthefish`, `omf new plugin MyPlugin`).
- `omf.cli.update` / `omf.cli.install`: read `$status` of an unrelated command and reported success after failures.
- `omf.packages.remove`: `test` crashed on a missing/empty theme file; only the first of two package copies was removed.
- `omf.check.fish_prompt`: a plain `fish_prompt.fish` (what `fish_config` writes) was reported healthy, so `omf doctor` never caught it.
- `omf.check.version`: `sort -n` compared only the first field (`1.10.0 < 1.9.5`).
- `omf.bundle.add`: unanchored grep skipped `bass` when `bassx` existed. `omf.bundle.remove`: deleted the bundle before rewriting it.
- `omf.index.path` / `omf.index.update`: empty `XDG_CACHE_HOME` resolved the index to `/omf` and the cleanup `rm -rf` ran there.
- `omf.index.repositories`: whole-line fixed-string matching, file guards, `mkdir`.
- `omf.cli.theme`: `$_` is empty on fish 4, mangling the usage line.
- `omf.version`: chopped the first character off a commit hash when no tag was reachable; prints `unknown` for tarball installs.
- **Messages vanished without a terminal**: `echo (omf::em)"text"(omf::off)` is a cartesian product, and `set_color` prints nothing when TERM is unset/dumb (CI, `ssh host cmd`), so the whole argument disappeared — "Installing package", "successfully installed", every error message and `omf list` headings were silently dropped on fish 3.2–4.0. The colour helpers now always yield one element.
- `omf list` / `omf theme` exited 127 where `column` is not installed (Alpine); falls back to plain output.

**Environment report**

```
[...]

PR #968, opened 2026-08-27
ci: test fish 4.1–4.8, add lint job, refresh runners

mhenrixon:
# Description

Stacked on #969 (includes #966, #969). Installs git in the Alpine-based `ohmyfish/fish` images for fish ≥ 4.1 *before* checkout — without it, `actions/checkout` silently falls back to a tarball and the tests run on a checkout with no `.git`.

- Linux matrix extended from 4.0.0 to every 4.x minor up to 4.8.0 (`ohmyfish/fish` images exist for all of them).
- macOS: `macos-14` is deprecated and `macos-latest` already aliases `macos-26`, so run on `macos-15` and `macos-26`.
- New lint job: `fish --no-execute` on every `.fish` file and `bin/install` (one file per call — fish only checks its first argument), plus verification of `bin/install.sha256`.
- `actions/checkout` v4 → v7; `permissions`, `concurrency`, `timeout-minutes`.
- `Dockerfile` pinned fish 2.5.0, below the installer's 3.0.0 minimum, so it could not build.

**Environment report**

```
OS type:              Darwin (macOS 27)
Fish version:         fish, version 4.8.1
Git version:          git version 2.55.0
Git core.autocrlf:    no
Checking for a sane environment...
Your shell is ready to swim.
```

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] New and existing tests pass locally with my changes

https://claude.ai/code/session_01CbqrmCn2mDNML6uRTaZVEw

PR #949, opened 2025-01-06
Fix #897: "Install aborted: /mingw64/bin/git is Git for Windows which is not supported"

Piker98988:
# Description

Commented `assert_cmds` and turned the abort message into a warning. Clarified how to disable `core.autocrlf`. 

Fixes #897

**Environment report**

```
Oh My Fish version:   7
OS type:              Linux
Fish version:         fish, version 3.1.0
Git version:          git version 2.26.0
Git core.autocrlf:    no
Checking for a sane environment...
Your shell is ready to swim.
```

# Checklist:

- [x] My code follows the [style guidelines](https://github.com/oh-my-fish/oh-my-fish/blob/master/CONTRIBUTING.md#code-style) of this project
- [x] I have performed a self-review of my own code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] New and existing tests pass locally with my changes 
- [x] I have updated the SHA256 checksum for the installation script

PR #970, opened 2026-08-27
fix(install): crash paths, URL mangling, empty XDG vars, Git for Windows

### LINKED ISSUE #897: Install aborted: /mingw64/bin/git is Git for Windows which is not supported.
I don't know why this is happening, but for some reason, I can't install `omf` on Windows (Git Bash)! I installed `fishshell` easily but got stuck in this problem. I don't really know what information to provide (you can ask for it explicitly), but the information I can provide is the last terminal output:

```
TheCodeHeist@TheCodeHeistPC ~> curl -L https://get.oh-my.fish | fish
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0    199      0 --:--:-- --:--:-- --:--:--   199
100 19035  100 19035    0     0  13966      0  0:00:01  0:00:01 --:--:-- 13966
Install aborted: /mingw64/bin/git is Git for Windows which is not supported.
TheCodeHeist@TheCodeHeistPC ~>
```

### LINKED ISSUE #940: Installer does not check `$OMF_PATH` permissions

mhenrixon:
# Description

Stacked on #968 (includes #966, #969, #968). The installer now uses the same always-one-element colour helper for its highlighted messages. Adopts the closed-unmerged #949 (Git for Windows) and #954 (writable offline install).

- `report` was never defined: on the stable channel with no reachable tag the installer printed "Unknown command" and carried on with an unpinned checkout. Use `abort`.
- `sed 's/\.git//'` stripped the first `.git` anywhere in `OMF_REPO_URI` (`my.github.io` → `myhub.io`). Anchor to the end.
- `set -q XDG_DATA_HOME` is true for an empty variable, giving `/omf` as the install path and baking the same test into the generated `conf.d/omf.fish`. Use `test -n` in both places.
- Git for Windows: warn instead of aborting (#949, fixes #897). Include the fix command in the autocrlf message.
- `chmod -R u+w` after an offline copy (#954, fixes #940).
- `assert_cmds` now checks `git`, `grep`, `chmod`.
- The `which` wrapper printed nothing for regular files with BSD `readlink`, leaving abort messages blank on macOS.
- Accept `Yes`/`YES` at prompts; skip non-directories when triggering uninstall hooks.

Fixes #897, fixes #940

**Environment report**

```
OS type:              Darwin (macOS 27)
Fish version:         fish, version 4.8.1
Git version:          git version 2.55.0
Git core.autocrlf:    no
Your shell is ready to swim.
```

# Checklist:

- [x] My code follows the style guidelines of this project

https://claude.ai/code/session_01CbqrmCn2mDNML6uRTaZVEw


derekstavis:
This needs a rebase after the latest merge

